PR #360, opened 2025-11-13
chore: Reorganize the token provider

gtema:
Apply a newer provider structure to keep modules smaller and better
structured.


gtema:
@phiphi282 this would be a sample PR for reorganizing the provider even further. This is not completed and further token types should be split into the fernet submodule (impl of the MsgPackToken trait with the test like for application_credential)

PR #408, opened 2025-11-30
Fix/system assignments issue 143

konac-hamza:
@gtema  Following your review, I did not modify any SeaORM tables or models. 
I updated the implementation to follow the same logic used in the list method. 
Looking forward to your feedback. Thank you!


src/assignment/backend/sql/assignment/list.rs
```diff
@@ -160,49 +170,70 @@ pub async fn list_for_multiple_actors_and_targets(
             );
         }
         select = select.filter(cond);
+    } else {
+        include_system = true;
     }
+    // System query always filters by target_id = "system"
+    select_system = select_system.filter(db_system_assignment::Column::TargetId.eq("system"));
```
gtema: while it is true in practice hardcording it in the backend is wrong since it is a business logic and not a backend driver logic. The caller would expected to set the target_id to system always. This is also how the python keystone works

PR #426, opened 2025-12-06
feat: Include system assignments in the `list_for_multiple_actors_and_targets` method

konac-hamza:
Include system role assignments in the `list_for_multiple_actors_and_targets` method.


gtema:
You could have just squashed and rewritten commits, but it's ok. I know that certain workflows require elevated privileges and cannot be executed from forks, but I don't know how to deal with that so far

PR #439, opened 2025-12-09
feat: Implement token structure validation

konac-hamza:
I have extended UnscopeToken payload with Validators. Then i have implemented "Validate" trait in `Token` itself. 'validate' trait been called for UnscopeToken. Other scope implementations are waiting this PR verification #435

### DIFF
--- a/src/token/types/common.rs
+++ b/src/token/types/common.rs
@@ -12,21 +12,20 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-use base64::{Engine as _, engine::general_purpose::URL_SAFE};
+use base64::{Engine as _, engine::general_purpose::URL_SAFE_NO_PAD};
 use validator::ValidationError;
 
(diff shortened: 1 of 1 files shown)


src/token/types/unscoped.rs
```diff
@@ -25,22 +28,71 @@ use crate::token::{
...
+
     #[builder(default, setter(name = _methods))]
+    #[validate(length(min = 1))]
     pub methods: Vec<String>,
+
     #[builder(default, setter(name = _audit_ids))]
+    #[validate(custom(function = "validate_audit_ids"))]
     pub audit_ids: Vec<String>,
+
+    #[validate(custom(function = "validate_future_datetime"))]
```
gtema: we can't enforce this. The token passed by the user is parsed into this structure, and the token payload itself is valid. The resulting token would eventually be considered as expired, but it is still valid from the data point of view. Please remove any datetime validations.

gtema (approved):
great work, thanks

PR #454, opened 2025-12-14
feat: Implement create_role api and backend

konac-hamza:
Because of there is not any api calls to create role in DB i have tried to implement api and backend side together. Could you review commit,please?

### DIFF
--- a/src/api/v3/role/mod.rs
+++ b/src/api/v3/role/mod.rs
@@ -13,24 +13,23 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use axum::{
-    extract::{Json, Path, Query, State},
-    http::StatusCode,
+    extract::{Path, Query, State},
     response::IntoResponse,
 };
 use utoipa_axum::{router::OpenApiRouter, routes};
-use validator::Validate;
 
 use crate::api::auth::Auth;
 use crate::api::error::KeystoneApiError;
 use crate::assignment::AssignmentApi;
 use crate::keystone::ServiceState;
-use types::{Role, RoleCreate, RoleList, RoleListParameters, RoleResponse};
+use types::{Role, RoleList, RoleListParameters, RoleResponse};
 
+mod create;
 pub mod types;
 
 pub(crate) fn openapi_router() -> OpenApiRouter<ServiceState> {
     OpenApiRouter::new()
-        .routes(routes!(list, create))
+        .routes(routes!(list, create::create))
         .routes(routes!(show))
 }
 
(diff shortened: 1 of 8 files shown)


src/assignment/backend/sql/role/create.rs
```diff
@@ -0,0 +1,190 @@
...
+use crate::db::entity::role as db_role;
+
+/// Create a new role
+pub async fn create(
+    _conf: &Config,
+    db: &DatabaseConnection,
+    role: &RoleCreate,  // ← Using RoleCreate instead of Role
+) -> Result<Role, AssignmentDatabaseError> {
+    // Serialize extra field if present
+    let extra_json = role
```
gtema: look as an example at the user create json handling (https://github.com/openstack-experimental/keystone/blob/main/src/identity/backends/sql/user/create.rs#L56C1-L59C1) - there is no separate error, but it is covered by the general Serde error

src/api/v3/role/mod.rs
```diff
@@ -13,23 +13,25 @@
...
-use types::{Role, RoleList, RoleListParameters, RoleResponse};
+use types::{Role, RoleList, RoleListParameters, RoleResponse,RoleCreate};
 
 pub mod types;
 
 pub(crate) fn openapi_router() -> OpenApiRouter<ServiceState> {
     OpenApiRouter::new()
         .routes(routes!(list))
         .routes(routes!(show))
+        .routes(routes!(create))
```
gtema: `create` and `list` belong together since they are technically the same URL just with different methods. Look in other apis

src/assignment/backend/sql/role/create.rs
```diff
@@ -0,0 +1,190 @@
...
+use sea_orm::entity::*;
+
+use crate::assignment::backend::error::{db_err, AssignmentDatabaseError};
+use crate::assignment::types::role::{Role, RoleCreate};
+use crate::config::Config;
+use crate::db::entity::role as db_role;
+
+/// Create a new role
+pub async fn create(
+    _conf: &Config,
```
gtema: you can skip this completely. It is rarely needed in the backend driver. It is a legacy remaining in some of the first implementations

src/assignment/backend/sql/role/create.rs
```diff
@@ -0,0 +1,145 @@
...
+use uuid::Uuid;
+
+use crate::assignment::backend::error::{AssignmentDatabaseError, db_err};
+use crate::assignment::types::role::{Role, RoleCreate};
+use crate::db::entity::role as db_role;
+
+/// Create a new role
+pub async fn create(
+    db: &DatabaseConnection,
+    role: &RoleCreate, // ← Using RoleCreate instead of Role
```
gtema: mostly the backends create methods take creation struct by value and not by reference

src/assignment/backend/sql/role/create.rs
```diff
@@ -0,0 +1,150 @@
...
+use crate::assignment::backend::error::{AssignmentDatabaseError, db_err};
+use crate::assignment::types::role::{Role, RoleCreate};
+use crate::db::entity::role as db_role;
+
+/// Create a new role
+pub async fn create(
+    db: &DatabaseConnection,
+    role: RoleCreate, // ← Using RoleCreate instead of Role
+) -> Result<Role, AssignmentDatabaseError> {
+    let role_id = role.id.as_ref().ok_or_else(|| {
```
gtema: no, as I said previously this should fallback to the uuid and not throw an error

PR #458, opened 2025-12-16
feat: Extend identity route to list local users

konac-hamza:


konac-hamza:
> no, this is not the way how the provider should be implemented. You can extend the existing UserListParameters with the dedicated user type and pass it down to the regular listing function. Otherwise the provider interface is simply going to explode. Whether on the API level it make sense to expose it as a separate endpoint is a different question (which I am not able to answer as of now). For sure such new API will not go under the /v3

Could you review implemented `UserType` enum, please? If it looks ok, i'd like to start backend implementation.

src/identity/types/user.rs
gtema: This is fine

src/api/v3/user/types.rs
gtema: this is fine in the concept, but we are not going to expose this in the api v3 (so far). I suggest first you focus only on the backend/provider part. In the separate PR you can then start relying on that, but only in the v4 api (or at least we can have a separate discussion in the dedicated PR)

gtema: please remove this part of the change - no api changes for now

src/identity/backends/sql/local_user/list.rs
```diff
@@ -0,0 +1,110 @@
...
+    local_user as db_local_user, password as db_password,
+    prelude::{LocalUser, User as DbUser, UserOption},
+    user as db_user,
+};
+use crate::error::DbContextExt;
+use crate::identity::backends::error::IdentityDatabaseError;
+use crate::identity::types::*;
+
+/// Prepare the paginated query for listing local users.
+fn get_list_query(
```
gtema: I strongly discourage from adding this in this change - you make the change bigger and harder to review

gtema (changes requested):

gtema:
I have slightly reworked the method to easily implement conditional parallel tasks and reduce code duplication. Please update your PR to follow the approach

PR #524, opened 2026-01-19
feat: Implement grant deletion provider api

konac-hamza:


src/assignment/types/assignment.rs
```diff
@@ -293,3 +293,65 @@ pub enum RoleAssignmentTargetType {
     /// System ID.
     System,
 }
+
+#[derive(Builder, Clone, Debug, Deserialize, PartialEq, Serialize, Validate)]
+#[builder(build_fn(error = "BuilderError"))]
+#[builder(setter(strip_option, into))]
+pub struct AssignmentRevoke {
```
gtema: please implement the From<Assignment> for AssignmentRevoke. I think this is going to be more comfortable to use it this way (check the assignment and pass it down to the revoke), but honestly I doubt whether we need AssignmentRevoke at all (it's same except few optional fields) and we should just use the Assignment directly?

konac-hamza: During implementation, I also had doubts about the necessity of AssignmentRevoke.
Then I realized that the AssignmentCreate struct is almost the same as Assignment.
So I kept the same approach for AssignmentRevoke but use Assignment directly look more clear. I will update PR accordingly

gtema: create/update is a slightly different usecase. Usually they look very similar, but have differences (i.e. some attributes are not updatable)

src/api/v3/role_assignment/project/user/role/revoke.rs
```diff
@@ -0,0 +1,460 @@
...
+    ),
+    responses(
+        (status = 204, description = "Role revoked successfully"),
+        (status = 401, description = "Unauthorized"),
+        (status = 403, description = "Forbidden"),
+        (status = 404, description = "Grant not found", example = json!(KeystoneApiError::NotFound(String::from("id = 1"))))
+    ),
+    security(
+        ("X-Auth-Token" = [])),
+    tag="Role Assignment"
```
gtema: the openapi tag should match the one in the create assignment, it influences how endpoints are grouped

PR #639, opened 2026-03-23
feat: Add api to list user roles on project

### LINKED ISSUE #518: Add v3 api to list all roles that the user has on the project
https://docs.openstack.org/api-ref/identity/v3/index.html?expanded=check-whether-user-has-role-assignment-on-project-detail#list-role-assignments-for-user-on-domain

konac-hamza:


gtema (changes requested):

crates/keystone/src/api/v3/role_assignment/project/user/role/list.rs
```diff
@@ -0,0 +1,370 @@
...
+
+pub(super) async fn list(
+    Auth(user_auth): Auth,
+    Path((project_id, user_id)): Path<(String, String)>,
+    State(state): State<ServiceState>,
+) -> Result<impl IntoResponse, KeystoneApiError> {
+    let query_params = RoleAssignmentListParameters {
+        user_id: Some(user_id.clone()),
+        project_id: Some(project_id.clone()),
+        effective: Some(true),
```
gtema: In the meanwhile I have spotted the python keystone return the list of direct grants on those apis, and not the effective ones. We would need to adopt the existing methods that we have already.

crates/keystone/src/api/v3/role_assignment/project/user/role/list.rs
```diff
@@ -0,0 +1,370 @@
...
+
+        let response = openapi_router()
+            .layer(TraceLayer::new_for_http())
+            .with_state(state)
+            .as_service()
+            .oneshot(
+                Request::builder()
+                    .method("GET")
+                    .uri("/projects/project_id/users/user_id/roles")
+                    .header("x-auth-token", "foo")
```
gtema: you would need to adopt this slightly - in the meanwhile a quite massive rework landed that changed few things about authentication handling and the tests now also pass a mocked ValidatedSecurityContext as extension instead of expecting the token would be validated.

policy/project/user/role/list.rego
```diff
@@ -0,0 +1,27 @@
...
+allow if {
+    "admin" in input.credentials.roles
+}
+
+allow if {
+    "manager" in input.credentials.roles
+    assignment.project_role_domain_matches
+}
+
+allow if {
```
gtema: the user should NOT be able to read own roles when he doesn't have any role on the scope. Technically the whole rule should be removed. For now just repeat the 'check' policy

gtema (changes requested):
thanks for reviving the work here. Unfortunately some more fixes are necessary since in the meanwhile the code base run away slightly ;-)

gtema (approved):

PR #671, opened 2026-04-20
feat: Add basic healthcheck endpoint

### LINKED ISSUE #521: Implement healthcheck endpoint
In order to simplify deployment under the kubernetes we should provide certain healthcheck/ready endpoints

phiphi282:
Add a basic healthcheck endpoint that returns the health of important dependencies like the database, opa or raft.


gtema:
thanks @phiphi282 I allowed myself to extend your work a bit to have a bit different structure of the response allowing also WARN states so that there is a difference between /ready and /health (a.k.a. /live)

gtema (approved):

PR #742, opened 2026-05-31
core: return ActorHasNoRolesOnTarget for empty scoped roles

### LINKED ISSUE #514: ensure 401 is returned in /auth/tokens when the user has no roles on the scope
Currently the 500 is being returned

magdang:
## Description
This PR resolves an issue where requesting a scoped authentication token with an explicitly empty roles array (`[]`) causes an internal server crash (or incorrect context state tracking) instead of dropping cleanly into a 401 Unauthorized response.

### Changes Implemented
- **Core Types Logic Fix**: Modified the `fully_resolved` evaluation inside `crates/core-types/src/auth.rs`.
- **Semantic Disambiguation**: Differentiated between `None` (roles have not yet been evaluated/populated from persistence) and `Some(roles) if roles.is_empty()` (roles were explicitly evaluated and the user holds zero assignments on the designated target).
- **Correct Error Trait Mapping**: Updated the empty role array match arm to natively trigger an `ActorHasNoRolesOnTarget` authentication error rather than falling back to `SecurityContextNotResolved`.
- **Test Suite Alignment**: Synchronized the underlying context assertion expectations inside `make_auth_result_project` and `make_trust_with_roles` submodule unit tests.

Closes #514

### DIFF
--- a/crates/core-types/src/auth.rs
+++ b/crates/core-types/src/auth.rs
@@ -769,13 +769,7 @@ impl SecurityContext {
             .authorization
             .as_ref()
             .ok_or(AuthenticationError::SecurityContextNotResolved)?;
-        // Unscoped with no roles is valid. Scoped with no roles OR empty roles list is
-        // not.
-        if !matches!(authz.scope, ScopeInfo::Unscoped)
-            && authz.roles.as_ref().is_none_or(|r| r.is_empty())
-        {
-            return Err(AuthenticationError::SecurityContextNotResolved);
-        }
+
         Ok(())
     }
 }
@@ -2135,24 +2129,6 @@ mod tests {
     }
 
     #[test]
-    fn test_fully_resolved_scoped_none_roles() {
-        let ctx = make_auth_result_project(make_principal("uid"), make_project(), None);
-        assert!(matches!(
-            ctx.fully_resolved(),
-            Err(AuthenticationError::SecurityContextNotResolved)
-        ));
-    }
-
-    #[test]
-    fn test_fully_resolved_scoped_empty_roles() {
-        let ctx = make_auth_result_project(make_principal("uid"), make_project(), Some(vec![]));
-        assert!(matches!(
-            ctx.fully_resolved(),
-            Err(AuthenticationError::SecurityContextNotResolved)
-        ));
-    }
-
-    #[test]
     fn test_fully_resolved_scoped_with_roles() {
         let ctx = make_auth_result_project(
             make_principal("uid"),
@@ -2169,27 +2145,9 @@ mod tests {
     }
 
     #[test]
-    fn test_fully_resolved_system_none_roles() {
-        let ctx = make_auth_result_system(make_principal("uid"), None);
-        assert!(matches!(
-            ctx.fully_resolved(),
-            Err(AuthenticationError::SecurityContextNotResolved)
-        ));
-    }
-
-    #[test]
     fn test_fully_resolved_domain_with_roles() {
         let ctx = make_auth_result_domain(make_principal("uid"), Some(vec![admin_role()]));
         assert!(ctx.fully_resolved().is_ok());
-    }
-
-    #[test]
-    fn test_fully_resolved_domain_none_roles() {
-        let ctx = make_auth_result_domain(make_principal("uid"), None);
-        assert!(matches!(
-            ctx.fully_resolved(),
-            Err(AuthenticationError::SecurityContextNotResolved)
-        ));
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


crates/core-types/src/auth.rs
```diff
@@ -761,20 +761,25 @@ impl SecurityContext {
...
+    /// * `Err(AuthenticationError::ActorHasNoRolesOnTarget)` if a scoped
+    ///   authorization has an explicitly empty role set.
     #[must_use = "discarding the result allows incomplete contexts to pass through"]
     pub fn fully_resolved(&self) -> Result<(), AuthenticationError> {
         self.validate()?;
         let authz = self
             .authorization
             .as_ref()
             .ok_or(AuthenticationError::SecurityContextNotResolved)?;
-        // Unscoped with no roles is valid. Scoped with no roles OR empty roles list is
```
gtema, line 772: can you please leave these comments lines in? they are relatively helpful explaining the match condition

gtema (changes requested):
thanks a lot for starting working on it. I totally get the the issue description is not detailed enough. In order to be able to mark issue as complete we should not only replace one error with another, but to actually introduce a test verifying that the status code is really 401. This can be achieved in 2 different ways:
- a new unittest in `crates/keystone/src/api/v3/auth/token/create.rs` simulating the error being thrown
- a new test in `tests/api/...` that creates project/user and tries to authenticate new user to that new project with role being assigned to the user on the project expecting 401

Actually another issue in the current flow is that the function you adopted (fully_resolved) is only invoked when user is already authenticated (in the crates/core/src/api/auth.rs) and not when the user tries to authenticate which is what we try to verify here.

I would suggest moving the check (the whole match) into `crates/keystone/core/src/auth.rs` at the bottom of the `new_with_roles` (right between the if block that calculates roles and that returns the ValidatedSecurityContext. A new unittest would need to be created in this module testing that the particular exception is returned.
In addition you would need also to add a new unittest into crates/keystone/src/api/v3/auth/token/create.rs. Here the "expect_issue_token_context" mock method would return the ActorHasNoRolesOnTarget error converted into the TokenProviderError::Authorization. Since during the conversion of the TokenProviderError to the final API response an unauthorized error is being returned (`crates/api-types/src/error_conv.rs` lines 249 and 81) it should result exactly in the 401 status code coming back from the api

magdang:
I have updated the code exactly as you suggested. I moved the role validation check out of fully_resolved and placed it at the bottom of new_for_scope. I also added the new unit tests in this module to verify the error is thrown, as well as the new API unit test in token/create.rs to mock the error and ensure it returns a 401 status code. Finally, I removed the old tests in core-types that were still checking the old logic, and the workspace tests are now passing successfully.

gtema (approved):

crates/core/src/auth.rs
```diff
@@ -2269,4 +2314,144 @@ mod tests {
...
+        // Same strict predicate as Test 2 — the provider IS called but returns
+        // an empty list, triggering the ActorHasNoRolesOnTarget guard.
+        let mut assignment_mock = MockAssignmentProvider::default();
+        assignment_mock
+            .expect_list_role_assignments()
+            .withf(|_, q: &RoleAssignmentListParameters| {
+                q.user_id.as_deref() == Some(uid)
+                    && q.project_id.as_deref() == Some(pid)
+                    && q.effective == Some(true)
+                    && q.include_names == Some(false)
```
gtema: include_names should be true

crates/core/src/auth.rs
```diff
@@ -2269,4 +2314,144 @@ mod tests {
...
+        // Strict predicate: must be called once for this user+project combination
+        // with the exact flags used by resolve_project_default_roles.
+        let mut assignment_mock = MockAssignmentProvider::default();
+        assignment_mock
+            .expect_list_role_assignments()
+            .withf(|_, q: &RoleAssignmentListParameters| {
+                q.user_id.as_deref() == Some(uid)
+                    && q.project_id.as_deref() == Some(pid)
+                    && q.effective == Some(true)
+                    && q.include_names == Some(false)
```
gtema: include_names should be true

crates/core/src/auth.rs
```diff
@@ -1708,6 +1716,43 @@ mod tests {
...
+        let uid = "uid";
+        let pid = "pid";
+        let mut assignment_mock = MockAssignmentProvider::default();
+        assignment_mock
+            .expect_list_role_assignments()
+            .withf(|_, q: &RoleAssignmentListParameters| {
+                q.user_id.as_deref() == Some(uid)
+                    && q.project_id.as_deref() == Some(pid)
+                    && q.effective == Some(true)
+                    && q.include_names == Some(false)
```
gtema: include_names should be true

gtema (changes requested):
small change landed in the meanwhile that requires you to do minor adjustment to the tests

crates/core/src/auth.rs
magdang: I have made the adjustments.

gtema (approved):

gtema:
The failure is unrelated and is caused by upstream openstack_cli that just got released. I am looking into it

PR #751, opened 2026-06-04
Add integration tests for role assignments with same role on multiple scopes (Issue #513)

### LINKED ISSUE #513: Add integration test for listing role assignments for particular user
The API test for trying to rescope into one of the projects it has access to caught few issues:
- listing role assignments by only a user was mixing up roles assigned to system and project due to the use of HashMap by role_id (it is absolutely normal to have multiple assignments for the same role on different targets)
- corrupted list was returned in the /auth/projects from the api and internal filter (that only lets GroupProject, UserProject assignments) resulted in the empty list
- empty list went into the list of the projects returning all projects
- auth into the scope that the user does not have roles on returns 500

Some of those errors should have been caught already on the integration test:

- assign a user the same role on different projects (including system)
  - list role assignments passing only actor_id
  - list role assignments passing only role_id

MohidSheraz:
Closes #513

Added two integration tests to catch the bugs described in issue #513:

- test_list_role_assignments_by_user_same_role_multiple_scopes
  - Assigns same role to user on 2 projects + system
  - Verifies listing by user_id returns all 3 assignments

- test_list_role_assignments_by_role_id_same_role_multiple_scopes  
  - Assigns same role to user on 2 projects + system
  - Verifies listing by role_id returns all 3 assignments

Both tests pass successfully.

### DIFF
--- a/tests/integration/src/assignment/grant/list.rs
+++ b/tests/integration/src/assignment/grant/list.rs
@@ -200,3 +200,98 @@ async fn test_list_user_roles() -> Result<()> {
     );
     Ok(())
 }
+#[traced_test]
+#[tokio::test]
+async fn test_list_role_assignments_by_user_same_role_multiple_scopes() -> Result<()> {
+    let (state, _) = get_state().await?;
+
+    let domain = create_domain!(state)?;
+    let project_1 = create_project!(state, domain.id.clone())?;
+    let project_2 = create_project!(state, domain.id.clone())?;
+    let user = create_user!(state, domain.id.clone())?;
+    let role = create_role!(state)?;
+
+    for assignment in [
+        AssignmentCreate::user_project(&user.id, &project_1.id, &role.id, false),
+        AssignmentCreate::user_project(&user.id, &project_2.id, &role.id, false),
+        AssignmentCreate::user_system(&user.id, "all", &role.id, false),
+    ] {
+        state
+            .provider
+            .get_assignment_provider()
+            .create_grant(&state, assignment)
+            .await?;
+    }
+
+    let assignments = state
+        .provider
+        .get_assignment_provider()
+        .list_role_assignments(
+            &state,
+            &RoleAssignmentListParametersBuilder::default()
+                .user_id(user.id.clone())
+                .build()?,
+        )
+        .await?;
+
+    assert_eq!(assignments.len(), 3);
+
+    let target_ids: BTreeSet<String> = assignments.iter().map(|a| a.target_id.clone()).collect();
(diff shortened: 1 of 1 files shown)


tests/integration/src/assignment/grant/list.rs
```diff
@@ -200,3 +200,140 @@ async fn test_list_user_roles() -> Result<()> {
     );
     Ok(())
 }
+
+// Add these two tests at the BOTTOM of:
```
gtema: this is likely the AI generated content and is making no sense here

tests/integration/src/assignment/grant/list.rs
```diff
@@ -200,3 +200,140 @@ async fn test_list_user_roles() -> Result<()> {
...
+    // Verify both project targets are present
+    let target_ids: BTreeSet<String> = assignments.iter().map(|a| a.target_id.clone()).collect();
+    assert!(
+        target_ids.contains(&project_1.id),
+        "Assignment on project_1 should be present"
+    );
+    assert!(
+        target_ids.contains(&project_2.id),
+        "Assignment on project_2 should be present"
+    );
```
gtema: please also add explicit validation that the role is granted on the system scope

tests/integration/src/assignment/grant/list.rs
```diff
@@ -200,3 +200,140 @@ async fn test_list_user_roles() -> Result<()> {
...
+        user_assignments.iter().map(|a| a.target_id.clone()).collect();
+    assert!(
+        target_ids.contains(&project_1.id),
+        "Assignment on project_1 should be present when filtering by role_id"
+    );
+    assert!(
+        target_ids.contains(&project_2.id),
+        "Assignment on project_2 should be present when filtering by role_id"
+    );
+
```
gtema, line 295: also here should verify that a role is present on the system scope, not only the project

gtema (changes requested):
Thanks.
apart of the other changes the commit message would need to be changed (use `git commit --amend` and specify a new message following the conventional commit summary (e.g., `chore: Test role assignments with same role on multiple scopes`)

tests/api/src/auth/project.rs
```diff
@@ -57,3 +57,80 @@ async fn test_list_user_projects() -> Result<()> {
     assert!(!projects.is_empty());
     Ok(())
 }
+
+#[tokio::test]
```
gtema: this change does not belong to this pull request

gtema (changes requested):

gtema (approved):

PR #752, opened 2026-06-04
Add API test for /auth/projects with empty role assignments (Issue #515)

### LINKED ISSUE #515: Add test to the /auth/projects with empty assignments list being returned
As described in the parent issue an explicit test should be added to the api implementation with an empty assignments list being returned from the provider

MohidSheraz:
Closes #515

Added an API test to verify that when a user has no role assignments, 
the /auth/projects endpoint returns an empty list instead of all projects.

This is a regression test for the bug described in the parent issue #513.

Test: test_auth_projects_empty_when_no_roles
- Attempts to list projects for a user without authentication/roles
- Verifies either 401 (unauthorized) or empty project list
- Should NOT return all projects in the system

### DIFF
--- a/tests/api/src/auth/project.rs
+++ b/tests/api/src/auth/project.rs
@@ -57,3 +57,85 @@ async fn test_list_user_projects() -> Result<()> {
     assert!(!projects.is_empty());
     Ok(())
 }
+
+
+#[tokio::test]
+async fn test_auth_projects_returns_empty_for_user_without_roles() -> Result<()> {
+    // Regression test for issue #515:
+    // When a user has no role assignments, /auth/projects should return
+    // an empty list, not all projects in the system.
+    
+    let mut admin_client = TestClient::default()?;
+    admin_client.auth_admin().await?;
+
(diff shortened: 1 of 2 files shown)


tests/api/src/auth/project.rs
```diff
@@ -57,3 +57,85 @@ async fn test_list_user_projects() -> Result<()> {
...
 }
+
+
+#[tokio::test]
+async fn test_auth_projects_returns_empty_for_user_without_roles() -> Result<()> {
+    // Regression test for issue #515:
+    // When a user has no role assignments, /auth/projects should return
+    // an empty list, not all projects in the system.
+    
+    let mut admin_client = TestClient::default()?;
```
gtema, line 68: Please do not use `TestClient` and instead use to openstack_sdk AsyncOpenStack (https://github.com/openstack-experimental/keystone/blob/main/tests/api/src/role/imply/check.rs#L30C16-L30C63)

gtema (changes requested):

PR #753, opened 2026-06-04
fix: Enforce minimum range boundaries for security compliance (fixes #746)

### LINKED ISSUE #746: config.security_compliance.password_expires_days must be >0
Some fields in the Config crate should have range checks (@crates/config/src/security_compliance.rs:password_expires_days should have min=1). Constraints themselves should be copied (in the security compliance case) from https://opendev.org/openstack/keystone/src/branch/master/keystone/conf/security_compliance.py#L66  Since there is no possibility to implement this easily with serde the easier approach is to use `validate` crate which is already used in the project to set additional range constraints.
The config manager should be forcibly call self.validate to enforce the checks.

magdang:
This patch introduces runtime configuration validation rules ensuring that specific security compliance settings (like 'password_expires_days', 'disable_user_account_days_inactive', etc.)  mirroring openstack/keystone python implementation.

The config crate has been updated to use the validator crate, and the ConfigManager now invokes validation on startup and file reloads.

Closes #746

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3873,6 +3873,7 @@ dependencies = [
  "tracing",
  "url",
  "url-macro",
+ "validator",
 ]
 
 [[package]]
--- a/crates/config/Cargo.toml
+++ b/crates/config/Cargo.toml
@@ -24,6 +24,7 @@ tokio = { workspace = true, features = ["fs", "rt", "sync", "time"] }
 tracing.workspace = true
 url = { workspace = true, features = ["serde"] }
 url-macro.workspace = true
+validator = { workspace = true, features = ["derive"] }
 
 [dev-dependencies]
 serde_json.workspace = true
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -54,6 +54,7 @@ use notify::{RecommendedWatcher, RecursiveMode, Watcher};
 use serde::Deserialize;
 use tokio::sync::RwLock;
 use tracing::error;
+use validator::Validate;
 
 mod application_credentials;
 mod assignment;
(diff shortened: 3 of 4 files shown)


gtema (changes requested):
Looks good. 
Would you mind adding a unittest (on the lib level and not under the security_compliance section) that explicitly tries to parse a config that would fail the validation ensuring the error is actually thrown. This way we can be sure that we prevent for accidental removing of the validation. I would be also interested to just to know whether the error message that the user will see will contain explanation of what is wrong or whether it is just a "failed validation" message with no further details. The scope is not to test validation of every single field, but to ensure the error message contain information useful to the user.

magdang:
Thanks for the review!

I have added the unit test at the lib level (crates/config/src/lib.rs:test_invalid_security_compliance_validation). It writes an invalid configuration block containing 0 values to ensure that configuration parsing gets rejected as it should.

Regarding the error visibility: yes, the validator error message is implemented and provides clear context to the user. 

Note:
During the CI run, you might notice that the integration test test_revoke_user_project_grant_auth_impact fails with a TokenRevoked error.

After looking at the trace logs, I found that this is a preexisting quirk in the test fixture surfaced by my new validation constraints. The test mocks an ApplicationCredential token with an issued_at timestamp hardcoded to the Unix Epoch (1970-01-01T00:00:00Z).

Because the configuration now strictly validates and enforces non-zero security lifetimes, the revocation engine evaluates this token against the RevokeProvider's default expiration_buffer (1800 seconds). Seeing that the token is effectively 56 years old, the system correctly flags it as invalid/revoked against recent database security events before the test can proceed.

Proposed Solution:
This can be easily resolved by updating the test fixture in tests/integration/src/assignment/grant/revoke.rs to use a current timestamp for issued_at instead of the Epoch date.

Let me know if you would like me to include that test update in this PR, or if you prefer to handle it in a separate issue!

gtema (approved):
Great, thanks

PR #761, opened 2026-06-05
feat: Add region CRUD to catalog SQL driver

### LINKED ISSUE #723: Extend catalog-sql driver with region CRUD commands

mashawes:
feat: add region CRUD to catalog-driver-sql (#723)

Extends the catalog-sql driver with region CRUD, following the existing SQL-driver patterns.

- Adds Region / RegionCreate / RegionUpdate / RegionListParameters types in core-types
- Adds a region module to catalog-driver-sql (list/get/create/update/delete) with entity/domain conversions
- Wires the operations into the CatalogBackend trait and SqlBackend impl
- Adds a RegionNotFound error variant
- Unit tests for each operation using MockDatabase

Reads mirror the existing service/endpoint modules. Create/update/delete follow the patterns used in other SQL drivers.

Closes #723

Note: this contribution was developed with AI assistance

### DIFF
--- a/crates/core-types/src/catalog/region.rs
+++ b/crates/core-types/src/catalog/region.rs
@@ -1,0 +1,90 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+use std::collections::HashMap;
+
+use derive_builder::Builder;
+use serde_json::Value;
+use validator::Validate;
+
+use crate::error::BuilderError;
+
+#[derive(Builder, Clone, Debug, Default, PartialEq, Validate)]
+#[builder(build_fn(error = "BuilderError"))]
+#[builder(setter(strip_option, into))]
+pub struct Region {
+    /// The region description.
+    #[builder(default)]
+    #[validate(length(max = 255))]
+    pub description: Option<String>,
+
+    /// Additional region properties.
+    #[builder(default)]
+    pub extra: Option<Value>,
+
+    /// The ID of the region.
+    #[validate(length(min = 1, max = 255))]
+    pub id: String,
+
+    /// The ID of the parent region, when this region is nested under another.
+    #[builder(default)]
+    #[validate(length(max = 255))]
+    pub parent_region_id: Option<String>,
+}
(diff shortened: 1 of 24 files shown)


crates/core-types/src/catalog/region.rs
```diff
@@ -0,0 +1,72 @@
...
+
+use derive_builder::Builder;
+use serde_json::Value;
+
+use crate::error::BuilderError;
+
+#[derive(Builder, Clone, Debug, Default, PartialEq)]
+#[builder(build_fn(error = "BuilderError"))]
+#[builder(setter(strip_option, into))]
+pub struct Region {
```
gtema, line 26: A (maybe not written yet convention) is to sort struct fields alphabetically (and separate them with empty line). This helps heavily when structures are having many fields and a field need to be added or modified).

crates/core-types/src/catalog/region.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// The ID of the region.
+    pub id: String,
```
gtema: We would need to add validator crate macros for the fields (typically only length). You can take the limits themselves from https://opendev.org/openstack/keystone/src/branch/master/keystone/catalog/backends/sql.py#L33 and see how it is usually defined in https://github.com/openstack-experimental/keystone/blob/main/crates/core-types/src/application_credential/application_credential.rs#L55

crates/core/src/catalog/backend.rs
```diff
@@ -96,4 +96,78 @@ pub trait CatalogBackend: Send + Sync {
         state: &ServiceState,
         enabled: bool,
     ) -> Result<Vec<(Service, Vec<Endpoint>)>, CatalogProviderError>;
+
```
gtema, line 97: another convention is to keep method definitions in traits and impls sorted alphabetically. It has the same reason - easier further maintenance and review by humans

gtema (changes requested):
mostly good, but few refinements are still necessary

mashawes:
Thanks for the review! Addressed all three points. Sorted struct fields and trait/impl methods alphabetically with blank lines between fields, and added validator length limits with max 255 characters taken from the Python region model. Ready for another look.

gtema (changes requested):
Good. Now adding these new methods to the provider API itself is missing to make PR as a functionally complete (crates/core/src/catalog/{mod, provider_api, mock, service}.rs). Don't forget the the implementation in service.rs (the real provider logic) would need to invoke the 'validate' function to perform the validation that you have added as early in the function as possible. At the moment this is potentially not happening in every single provider, since the validation has been added later.
Also we would need to add integration tests for the provider API. Ask your agent to create sophisticated crud tests in integration test suite to cover as much as possible. With this the PR would be complete and the API layer would be added separately.

mashawes:
Addressed the review. Wired the region methods through the provider API including the service and mock layers, added a Validation error variant with early validate() calls in create, update, and list, and added integration tests covering the full CRUD flow plus parent filtering, not found, and over-length validation cases. Thanks!

gtema (approved):

PR #766, opened 2026-06-07
feat: Extend the project creation api in v4

konac-hamza:
- Added `ProjectCreate` type with an optional `ProjectID` field and remaining v3 project types in the `api-types` crate
- Added a UUID validator that enforces UUID v4 format on incoming project ID parameters
- All projects endpoints except `create` now delegate v3 routing to a public method defined in the `keystone` crate


crates/api-types/src/v4/project.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// If set to true, project is enabled. If set to false, project is
+    /// disabled. The defaults is `true`.
+    #[cfg_attr(feature = "builder", builder(default = "crate::default_true()"))]
+    pub enabled: bool,
+
+    /// Additional project properties.
+    #[cfg_attr(feature = "builder", builder(default))]
+    #[cfg_attr(feature = "openapi", schema(inline, additional_properties))]
+    #[serde(flatten)]
+    //pub extra: ExtraFields,
```
gtema: please remove this commented line. I know it is the copy, just no sense to copy dead code over

crates/keystone/src/api/v4/project/create.rs
```diff
@@ -0,0 +1,270 @@
+use axum::{
```
gtema, line 16: please add a license header as in other files

crates/api-types/src/v4/project.rs
```diff
@@ -0,0 +1,120 @@
...
+#[cfg_attr(feature = "validate", derive(validator::Validate))]
+pub struct ProjectCreateRequest {
+    /// Project object.
+    #[cfg_attr(feature = "validate", validate(nested))]
+    pub project: ProjectCreate,
+}
+
+// The Validator that check format of the project id if it given by user.
+#[cfg(feature = "validate")]
+fn validate_uuid(id: &str) -> Result<(), ValidationError> {
```
gtema, line 115: we are currently discussing this in https://review.opendev.org/c/openstack/keystone-specs/+/983440 One big problem is that we need to ensure it is a dash-less uuid. Otherwise we must reject it as bad request

crates/keystone/src/api/v4/project/create.rs
```diff
@@ -0,0 +1,270 @@
...
+use crate::api::v3::project::types::ProjectResponse;
+use crate::api::v4::project::types::ProjectCreateRequest;
+use crate::keystone::ServiceState;
+use crate::resource::ResourceApi;
+
+/// Create project.
+///
+/// Creates a project, where the project may act as a domain.
+#[utoipa::path(
+    post,
```
gtema, line 36: we definitely need to add `operation_id` param otherwise the openapi spec clashes

crates/api-types/src/v4/project.rs
```diff
@@ -0,0 +1,120 @@
...
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[cfg_attr(
+    feature = "builder",
+    derive(derive_builder::Builder),
+    builder(
+        build_fn(error = "crate::error::BuilderError", validate = "Self::validate"),
+        setter(strip_option, into)
+    )
+)]
+#[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
```
gtema: additional attribute `..., derive(utoipa::ToSchema), schema(as ProjectCreateV4)` is necessary to separate them in the openapi

crates/api-types/src/v4/project.rs
```diff
@@ -0,0 +1,120 @@
...
+        if self.parent_id.is_some() && self.is_domain.is_some_and(|x| x) {
+            return Err("project cannot specify `parent_id` when `is_domain` is true".to_string());
+        }
+        Ok(())
+    }
+}
+
+/// New project creation request.
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
```
gtema: here you also need to "rename" the openapi schema with `schema(as ProjectCreateRequestV4)`

crates/keystone/src/api/v4/project/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+#[openapi(
+    tags(
+        (name="projects", description="Project API - v4 supports optional project_id"),
+    )
+)]
+pub struct ApiDoc;
+
+pub(super) fn openapi_router() -> OpenApiRouter<ServiceState> {
+    OpenApiRouter::new()
+        .routes(routes!(create::create)) // v4-specific with optional id
```
gtema, line 34: problem with that merge is that it technically overwrites the new v3 handler. The unittest is not able to catch this you just mock the response. You would be only capable capturing this in the api tests. I would like to ask you to add explicit new tests into tests/api/src/resource/<maybe_project_v4> testing specifying the id

gtema (changes requested):

crates/api-types/src/v4/project.rs
```diff
@@ -0,0 +1,125 @@
...
+    /// Project object.
+    #[cfg_attr(feature = "validate", validate(nested))]
+    pub project: ProjectCreate,
+}
+
+// The Validator that check format of the project id if it given by user.
+#[cfg(feature = "validate")]
+fn validate_uuid(id: &str) -> Result<(), ValidationError> {
+    // Dashless UUIDs must be exactly 32 hex characters, no hyphens
+    let is_dashless = id.len() == 32 && !id.contains('-');
```
gtema: this is not a sufficient validation: uuid regulates also which characters can be present. Most reasonable approach is to try to parse it as uuid (try_parse), get the 'simple' formatter and compare it matches the input

gtema (changes requested):

crates/api-types/src/v4/project.rs
konac-hamza: Thanks for your review :). I will fix it as soon as possible.

PR #773, opened 2026-06-08
feat: Add service CRUD to the catalog provider

### LINKED ISSUE #724: Extend catalog-sql driver with service CRUD commands

mashawes:
Extends the catalog with service CRUD, following the same pattern as region (#723).

- Adds ServiceCreate and ServiceUpdate types in core-types with validator length limits (id max 64, type and name max 255) and alphabetized fields.
- Adds create, update, and delete to the catalog-sql driver. Read, list, and get already existed. A service's name is stored inside the extra JSON blob so create and update merge it there.
- Wires the new methods through the CatalogApi provider including provider_api, service, mock, and mod, calling validate() early in create, update, and list.
- Adds integration tests under tests/integration/src/catalog/service/ covering create, get, list with type filter, update, and delete, plus not-found and over-length validation cases.

Closes #724

### DIFF
--- a/crates/core-types/src/catalog/service.rs
+++ b/crates/core-types/src/catalog/service.rs
@@ -38,15 +38,25 @@ pub struct Service {
     #[validate(length(min = 1, max = 64))]
     pub id: String,
 
-    /// The service name.
-    #[builder(default)]
-    #[validate(length(max = 255))]
-    pub name: Option<String>,
-
     /// The service type.
     #[builder(default)]
     #[validate(length(max = 255))]
     pub r#type: Option<String>,
+}
+
+impl Service {
+    /// Returns the service name, if set.
+    ///
+    /// The name is not a dedicated database column; it is stored as the `name`
+    /// key inside the `extra` JSON blob, so it is read back out from there
+    /// rather than being a field on the model itself.
+    pub fn name(&self) -> Option<String> {
+        self.extra
+            .as_ref()
+            .and_then(|extra| extra.get("name"))
+            .and_then(|name| name.as_str())
+            .map(ToString::to_string)
+    }
 }
 
 /// Parameters for creating a new service.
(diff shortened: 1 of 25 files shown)


crates/core-types/src/catalog/service.rs
```diff
@@ -12,40 +12,96 @@
...
 use serde_json::Value;
+use validator::Validate;
 
 use crate::error::BuilderError;
 
-#[derive(Builder, Clone, Debug, Default, PartialEq)]
+#[derive(Builder, Clone, Debug, Default, PartialEq, Validate)]
 #[builder(build_fn(error = "BuilderError"))]
 #[builder(setter(strip_option, into))]
 pub struct Service {
```
gtema, line 26: in the database model we have only id/enabled/type/extra. All other fields are captured as json under extra. As such we should not try to re-expose the "name" here separately since it is anyway an optional field. Please remove "name" field from models

gtema (changes requested):

crates/core-types/src/catalog/service.rs
```diff
@@ -12,40 +12,88 @@
...
     /// The ID of the service.
+    #[validate(length(min = 1, max = 64))]
     pub id: String,
-    /// The service type.
-    #[builder(default)]
-    pub r#type: Option<String>,
+
     /// The service name.
     #[builder(default)]
+    #[validate(length(max = 255))]
```
gtema: we can't have `name` in the response. I know it was already there, but now I see it is an error. We also cannot limit the name length.

gtema (changes requested):
Would you please also update the 'list' method to actually filter by 'name'. I was first confused why at all we have 'name' in the list parameters, but it is already there for python keystone so we also need it. In our case this would be a "post" filter applied after fetching data from the database. Please add corresponding unittests

gtema (approved):

PR #783, opened 2026-06-10
test: Auth projects returns empty for user without roles

### LINKED ISSUE #515: Add test to the /auth/projects with empty assignments list being returned

MohidSheraz:
Closes #515

Added API test to verify that when a user has no role assignments,
the /auth/projects endpoint returns an empty list instead of all
projects in the system.

Test: test_auth_projects_empty_for_user_without_roles
- Creates a user with no role assignments
- Authenticates as that user
- Calls /auth/projects
- Verifies the returned list is empty

### DIFF
--- a/tests/api/src/auth/project.rs
+++ b/tests/api/src/auth/project.rs
@@ -20,6 +20,10 @@ use openstack_keystone_api_types::v3::project::ProjectShort;
 use openstack_sdk::api::rest_endpoint_prelude::*;
 use openstack_sdk::{AsyncOpenStack, api::QueryAsync, config::CloudConfig};
 
+use crate::common::get_session_by_user_password;
+use crate::guard::ResourceGuard;
+use tracing_test::traced_test;
+
 #[derive(Clone, Debug)]
 struct AuthProjectsRequest {}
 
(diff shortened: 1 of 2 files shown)


gtema (changes requested):
- you again "reset" the state from you local checkout with older change. Do pull before you change

Following is what you need to add:

- establish regular openstack_sdk session for the user by name/password/domain_id
#tests/api/std/common.rs
```
/// Get AsyncOpenStack session for user by name, domain and password.
pub async fn get_session_by_user_password<U: AsRef<str>, D: AsRef<str>, P: AsRef<str>>(
    username: U,
    domain_id: D,
    password: P,
) -> Result<Arc<AsyncOpenStack>> {
    let config = CloudConfig {
        auth: Some(openstack_sdk::config::Auth {
            auth_url: Some(env::var("OS_AUTH_URL")?),
            username: Some(username.as_ref().to_string()),
            user_domain_id: Some(domain_id.as_ref().to_string()),
            password: Some(password.as_ref().into()),
            ..Default::default()
        }),
        ..Default::default()
    };
    Ok(Arc::new(AsyncOpenStack::new(&config).await?))
}
```

- The test itself would look like:
```
#[tokio::test]
#[traced_test]
async fn test_auth_projects_empty_for_user_without_roles() -> Result<()> {
    use crate::identity::user::create_user;
    use openstack_keystone_api_types::v3::user::UserCreateBuilder;
    use uuid::Uuid;

    let tc = Arc::new(AsyncOpenStack::new(&CloudConfig::from_env()?).await?);
    let name = format!("usr_{}", Uuid::new_v4().simple());
    let password = "TestPassword123!";

    // Create a user with no role assignments
    let guard = create_user(
        &tc,
        UserCreateBuilder::default()
            .name(&name)
            .domain_id("default")
            .enabled(true)
            .password(password)
            .build()?,
    )
    .await?;

    let user_client =
        get_session_by_user_password(&guard.name, &guard.domain_id, &password).await?;

    let projects = list_auth_projects(&user_client).await?;
    assert!(projects.is_empty());

    guard.delete().await?;
    Ok(())
}
```

Once you do it you will see the test fails throwing 403 (which is good since it find an error). The hint for analysis: crates/keystone/src/api/v3/auth/project.rs:L50 and policy/auth/project/list.rego

gtema:
as mentioned: the referred policy need to be modified. Actually it should be just 'default allow : = true' explaining in the comment that authentication extractor ensures only authenticated users can access the endpoint, but every authenticated user should be able to list accessible projects. Correspondingly the test file for the policy should be also updated.
The bootstrap command PR just landed which enables running api tests directly. Check for the changes in the CONTRIBUTOR.md for instructions

gtema:
-

PR #785, opened 2026-06-10
feat: Add endpoint CRUD to the catalog provider

### LINKED ISSUE #725: Extend catalog-sql driver with endpoint CRUD commands

mashawes:
Extends the catalog with endpoint CRUD, following the same pattern as service 

* Adds EndpointCreate and EndpointUpdate types in core-types with validator length limits (id and service_id max 64, interface and region_id max 255; url is left unbounded since it maps to a Text column) and alphabetized fields.
* Adds create, update, and delete to the catalog-sql driver. Read, list, and get already existed. Unlike service, an endpoint's interface, service_id, url, and region_id are real database columns, so no extra-blob merging is needed.
* Wires the new methods through the CatalogApi provider including provider_api, service, mock, and mod, calling validate() early in create, update, and list.
* Adds integration tests under tests/integration/src/catalog/endpoint/ covering create, get, list with interface filter, update, and delete, plus not-found and over-length validation cases.

Closes #725 

### DIFF
--- a/crates/core-types/src/catalog/endpoint.rs
+++ b/crates/core-types/src/catalog/endpoint.rs
@@ -12,16 +12,20 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+use std::collections::HashMap;
+
 use derive_builder::Builder;
 use serde_json::Value;
+use validator::Validate;
 
 use crate::error::BuilderError;
 
-#[derive(Builder, Clone, Debug, Default, PartialEq)]
+#[derive(Builder, Clone, Debug, Default, PartialEq, Validate)]
 #[builder(build_fn(error = "BuilderError"))]
 #[builder(setter(strip_option, into))]
 pub struct Endpoint {
     /// The ID of the endpoint.
+    #[validate(length(min = 1, max = 64))]
     pub id: String,
     /// The interface type, which describes the visibility of the endpoint.
     /// Value is:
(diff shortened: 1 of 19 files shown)


crates/core/src/catalog/service.rs
```diff
@@ -84,6 +101,22 @@ impl CatalogApi for CatalogService {
...
+    /// - `id`: The unique identifier of the endpoint.
+    ///
+    /// # Returns
+    /// A `Result` indicating success or a `CatalogProviderError`.
+    async fn delete_endpoint<'a>(
+        &self,
+        state: &ServiceState,
+        id: &'a str,
+    ) -> Result<(), CatalogProviderError> {
+        self.backend_driver.delete_endpoint(state, id).await
```
gtema: in the meanwhile one change landed that emits notifications upon resource create/delete/update. You will see it when you rebase your change. Please ensure you send notifications in your handlers

crates/core-types/src/catalog/endpoint.rs
```diff
@@ -51,14 +58,74 @@ pub struct Endpoint {
...
+
+    /// Additional endpoint properties.
+    pub extra: HashMap<String, Value>,
+
+    /// The ID of the endpoint. A UUID is generated when omitted.
+    #[validate(length(min = 1, max = 64))]
+    pub id: Option<String>,
+
+    /// The interface type (`public`, `internal`, or `admin`).
+    #[validate(length(max = 255))]
```
gtema: it's "8" and not "255" - see https://opendev.org/openstack/keystone/src/branch/master/keystone/catalog/backends/sql.py#L77

gtema (changes requested):
other than 2 small issues would be good to go

gtema (approved):

PR #787, opened 2026-06-11
fix: Align "extra" property handling

### LINKED ISSUE #777: update_XXX treatment of "extra" field cause full reset of unchanged properties
Current code in update operations usually uses something similar to 
```if let Some(extra) = service.extra {
        update_model.extra = Set(Some(serde_json::to_string(&extra)?));
    }
```
This causes all extra properties that are present in the current state, but are missing in the update to be fully reset. Python keystone merges dictionaries preserving them. Both approaches have disadvantages, but we need to be compatible with the python behavior and therefore need to change how the 'extra' field is handled in drivers. When the "value" of the key in the new dictionary is empty we should unset the key (if present) in the current state to enable unsetting of extra property fields.

mashawes:
Consistently make "extra" column to be `HashMap<String, Value>` for all resources supporting it on get/create/update repeating python keystone behavior or always overwriting the complete "extra" column with the passed properties

Closes #777

### DIFF
--- a/crates/core/src/catalog/service.rs
+++ b/crates/core/src/catalog/service.rs
@@ -21,6 +21,7 @@ use openstack_keystone_core_types::catalog::*;
 use openstack_keystone_core_types::events::{Event, EventPayload, Operation};
 
 use crate::catalog::{CatalogApi, CatalogProviderError, backend::CatalogBackend};
+use crate::db::merge_extra;
 use crate::keystone::ServiceState;
 use crate::plugin_manager::PluginManagerApi;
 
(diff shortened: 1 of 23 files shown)


gtema (changes requested):
The change on it's own is clean, there is however one critical issue: provider sends one value for update (which is also logged in traces), but drivers persist a different value. It also opens doors for situations that for 1 provider one drivers merges the data, while the second one misses it and overwrites the data. The same is valid across providers.

It must be responsibility of the provider to prepare the data, while the backend driver is only responsible for persistence.

crates/core/src/catalog/service.rs
```diff
@@ -352,9 +353,19 @@ impl CatalogApi for CatalogService {
         &self,
         state: &ServiceState,
         id: &'a str,
-        endpoint: EndpointUpdate,
+        mut endpoint: EndpointUpdate,
```
gtema: changing methods signatures should not be necessary. You already take the ownership of the structure so you can also mutate it (```let mut update = endpoint; if let ....```). This applies to all methods.

gtema (changes requested):

gtema (changes requested):
You know what, I am now looking at that and notice inconsistency: Project/Domain define extra as HashMap<String, Value> everywhere, while catalog resources as Option<Value> in response and HashMap<String, Value> in create/update. Let's change the types in the catalog provider to have also HashMap<String, Value> on the get responses to align for consistency
Further, I am sorry, but I have dug through python Keystone and found that it DOES NOT merge dicts. If we would do so we would introduce inconsistency in the behavior. https://opendev.org/openstack/keystone/src/branch/master/keystone/catalog/backends/sql.py#L186-L190 shows it takes whichever new properties came and overwrite it to 'extra' column. So let us have consistently `extra: HashMap<String, Value>` for get/create/update types on all resources supporting this column. Since python keystone actually crashes when the column is not at least '{}' we have implicit safety guards by eliminating the `None`

gtema (approved):

PR #806, opened 2026-06-15
feat: Add access rule CRD to appcred provider

### LINKED ISSUE #776: Add application_credentials.access_rule CRUD support in the provider
Extend the application_credentials driver and provider with CRD (no update) for access rules https://docs.openstack.org/api-ref/identity/v3/index.html#application-credentials

 [ ] - provider (create/list/show/delete)
 [ ] - integration test

mashawes:
Extends the application credential provider with access rule CRD (create, read, delete, no update), following the same layering as application credentials.

* Adds two error variants in core-types, AccessRuleNotFound and AccessRuleInUse. The AccessRule and AccessRuleCreate types already existed, so no new domain types were needed.
* Adds create, get, list, and delete to the appcred-sql driver under application_credential/access_rule/. Access rules are user-scoped, and the user-facing id maps to the external_id column. Delete refuses to remove a rule that is still attached to an application credential (AccessRuleInUse) rather than cascade-deleting it, since that would silently strip a live credential's restriction; this matches Python Keystone's behavior.
* Wires the new methods through the ApplicationCredentialApi provider including backend, provider_api, service, mock, and mod. The provider validates the request and generates the rule id when none is supplied, so the driver only has to persist the final value.
* Adds integration tests under tests/integration/src/application_credential/access_rule.rs covering the full create, get, list, and delete cycle, plus the not-found and in-use error cases.

Closes #776


crates/appcred-driver-sql/src/lib.rs
```diff
@@ -97,6 +97,77 @@ impl ApplicationCredentialBackend for SqlBackend {
...
+    /// A `Result` containing a `Vec` of `AccessRule` or an `Error`.
+    async fn list_access_rules<'a>(
+        &self,
+        state: &ServiceState,
+        user_id: &'a str,
+    ) -> Result<Vec<AccessRule>, ApplicationCredentialProviderError> {
+        application_credential::access_rule::list(&state.db, user_id).await
+    }
+
+    /// Delete a user's access rule by its ID.
```
gtema: methods should be sorted alphabetically in all traits and impls

crates/core/src/application_credential/backend.rs
```diff
@@ -71,4 +71,70 @@ pub trait ApplicationCredentialBackend: Send + Sync {
         state: &ServiceState,
         params: &ApplicationCredentialListParameters,
     ) -> Result<Vec<ApplicationCredential>, ApplicationCredentialProviderError>;
+
```
gtema, line 57: sort all methods alphabetically

gtema (changes requested):
please sort methods in all traits and impls that you modify alphabetically sorted. Otherwise looks fine

crates/appcred-driver-sql/src/application_credential/access_rule/delete.rs
```diff
@@ -0,0 +1,133 @@
...
+///
+/// # Parameters
+/// - `db`: The database connection.
+/// - `user_id`: The ID of the user owning the access rule.
+/// - `id`: The (external) ID of the access rule.
+///
+/// # Returns
+/// A `Result` containing `()` or an `Error` (`AccessRuleNotFound` if no such
+/// rule exists, `AccessRuleInUse` if it is still attached to a credential).
+pub async fn delete<U: AsRef<str>, I: AsRef<str>>(
```
gtema, line 46: please make 'db: &C' where C: ConnectionTrait you can find such code in some other places. Reason is that we would need to delete/insert in the transaction which requires this ConnectionTrait

crates/appcred-driver-sql/src/application_credential/access_rule/create.rs
```diff
@@ -0,0 +1,77 @@
...
+///
+/// # Parameters
+/// - `db`: The database connection.
+/// - `user_id`: The ID of the user owning the access rule.
+/// - `rule`: The access rule to create.
+///
+/// # Returns
+/// A `Result` containing the created `AccessRule` or an `Error`.
+pub async fn create<U: AsRef<str>>(
+    db: &DatabaseConnection,
```
gtema: make DatabaseConnection 'impl ConnectionTrait'

gtema (changes requested):
https://github.com/openstack-experimental/keystone/blob/main/crates/identity-driver-sql/src/local_user/create.rs#L44, https://github.com/openstack-experimental/keystone/blob/main/crates/identity-driver-sql/src/password/set.rs#L44 show that in some cases we need to invoke backend functions in the transaction context. For access rules it is definitely the case where create/delete operations would be bound to the parenting application_credential create/delete. This is a small interface change that will not even require unittest changes. Otherwise fine

crates/core/src/application_credential/service.rs
```diff
@@ -63,6 +63,33 @@ impl ApplicationCredentialService {
...
+    ) -> Result<AccessRule, ApplicationCredentialProviderError> {
+        let mut rule = rule;
+        rule.validate()?;
+        // The provider prepares the final data; the driver only persists it.
+        if rule.id.is_none() {
+            rule.id = Some(Uuid::new_v4().simple().to_string());
+        }
+        self.backend_driver
+            .create_access_rule(state, user_id, rule)
+            .await
```
gtema: one more finding: https://github.com/openstack-experimental/keystone/blob/main/crates/core/src/application_credential/service.rs#L118 shows us calling notification engine for create/update/delete operations. Please add it

crates/appcred-driver-sql/src/application_credential/access_rule/create.rs
```diff
@@ -0,0 +1,77 @@
...
+    user_id: U,
+    rule: AccessRuleCreate,
+) -> Result<AccessRule, ApplicationCredentialProviderError> {
+    db_access_rule::ActiveModel {
+        id: NotSet,
+        method: Set(rule.method),
+        path: Set(rule.path),
+        service: Set(rule.service),
+        external_id: Set(Some(rule.id.unwrap_or(Uuid::new_v4().simple().to_string()))),
+        user_id: Set(Some(user_id.as_ref().to_string())),
```
gtema: we should extend the AccessRuleCreate to have `user_id: String`

gtema (changes requested):
one, hopefully final thing. Please add "user_id" into the AccessRule core-types and correspondingly expose it from the database. It would be populated during creation and is definitely necessary in output. The AccessRule struct already contain a comment that in the DB user_id must be made not optional, but we can't do this in the keystone-rs project

gtema (approved):

PR #813, opened 2026-06-17
feat: Filter user listing by type

### LINKED ISSUE #234: Add user type listing param
Extend v4 user listing with user type parameter (or even dedicated apis) to explicitly list only local, only federated, only nonlocal users

mashawes:
Wires the type query parameter into user listing so callers can filter by local, federated, and nonlocal users.

The provider and driver already supported this: UserListParameters carries a user_type field, the SQL driver's list filters local/nonlocal/federated accordingly, and it has unit tests. What was missing was the API layer, nothing read the query parameter or passed it down.

* Adds a UserType enum and a user_type field (deserialized from the type query key) to the api-types UserListParameters, and maps it through to the provider type in the conversion. Because the v4 user listing re-uses the same api-types parameters and conversion, ?type= now works on both the v3 and v4 user list endpoints.
* Adds API tests for ?type=local on both v3 and v4 confirming the filter reaches the provider.

Closes #234

### DIFF
--- a/crates/api-types/src/v4/user.rs
+++ b/crates/api-types/src/v4/user.rs
@@ -13,7 +13,52 @@
 // SPDX-License-Identifier: Apache-2.0
 //! User resource types.
 
+use serde::{Deserialize, Serialize};
+
 pub use crate::v3::user::{
-    Federation, FederationProtocol, User, UserCreate, UserCreateRequest, UserList,
-    UserListParameters, UserOptions, UserResponse, UserUpdateRequest,
+    Federation, FederationProtocol, User, UserCreate, UserCreateRequest, UserList, UserOptions,
+    UserResponse, UserUpdateRequest,
 };
+
+/// User list parameters.
+///
+/// V4 extends the v3 listing with a `type` filter; this is intentionally a
+/// v4-only parameter (Python Keystone does not support it on v3).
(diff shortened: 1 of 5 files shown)


gtema (changes requested):
Associated issue lists this, but maybe not very precise: this must be done for the api v4 only without touching v3. We can't introduce this new parameter into v3 since python keystone do not support it

gtema (changes requested):
just a reminder: DCO failure, merge conflict

crates/api-types/src/v4/user.rs
```diff
@@ -13,7 +13,52 @@
...
+    /// Filter users by Domain ID.
+    #[cfg_attr(feature = "validate", validate(length(max = 64)))]
+    pub domain_id: Option<String>,
+
+    /// Filter users by Name.
+    #[cfg_attr(feature = "validate", validate(length(max = 255)))]
+    pub name: Option<String>,
+
+    /// Filter users by the federated unique ID.
+    #[cfg_attr(feature = "validate", validate(length(max = 64)))]
```
gtema: uniqueId is 255 chars long

gtema (changes requested):

gtema (approved):

gtema (approved):

PR #831, opened 2026-06-23
test: Add identity provider integration tests

### LINKED ISSUE #449: Implement integration tests for identity provider

mashawes:
Adds integration test coverage for the identity provider, following the existing test patterns and harness.

* Group CRUD under tests/integration/src/identity/group/: create, get (including not-found), list (including filtering by domain), and delete.
* User lifecycle: a delete test, a get_user_domain_id test, and a find_federated_user test (creating a user with a federated identity and looking it up by idp/unique id, plus the not-found case).
* Group membership under tests/integration/src/identity/user_group/: remove and set, bulk add_users_to_groups and remove_user_from_groups, and the expiring variants (add, remove, remove-many, set), complementing the existing add/list coverage.
* All tests use the existing harness (get_state, create_domain!, create_user!, create_group!) and the public provider API; no production code is changed. authenticate_by_password is exercised by the token integration tests.

Closes #449

### DIFF
--- a/tests/integration/src/identity/group/create.rs
+++ b/tests/integration/src/identity/group/create.rs
@@ -11,9 +11,24 @@
 // limitations under the License.
 //
 // SPDX-License-Identifier: Apache-2.0
+//! Test create group functionality.
 
-mod create;
-mod delete;
-mod get;
-mod list;
-mod update;
+use eyre::Result;
+use tracing_test::traced_test;
+
+use crate::common::get_state;
+use crate::{create_domain, create_group};
+
+#[tokio::test]
+#[traced_test]
+async fn test_create() -> Result<()> {
+    let (state, _tmp) = get_state().await?;
+    let domain = create_domain!(state)?;
+
+    let group = create_group!(state, domain.id.clone())?;
+
+    assert!(!group.id.is_empty(), "an id was generated");
+    assert!(!group.name.is_empty());
+    assert_eq!(group.domain_id, domain.id);
+    Ok(())
+}
--- a/tests/integration/src/identity/group/delete.rs
+++ b/tests/integration/src/identity/group/delete.rs
@@ -1,0 +1,57 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+//! Test delete group functionality.
+
+use eyre::Result;
+use tracing_test::traced_test;
+
+use openstack_keystone_core::auth::ExecutionContext;
+
+use crate::common::get_state;
+use crate::{create_domain, create_group};
+
+#[tokio::test]
+#[traced_test]
+async fn test_delete() -> Result<()> {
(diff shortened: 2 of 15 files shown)


tests/integration/src/identity/group/create.rs
```diff
@@ -0,0 +1,52 @@
...
+use crate::create_domain;
+
+#[tokio::test]
+#[traced_test]
+async fn test_create() -> Result<()> {
+    let (state, _tmp) = get_state().await?;
+    let domain = create_domain!(state)?;
+    let name = Uuid::new_v4().to_string();
+
+    let group = state
```
gtema: let's please use "create_user!", "create_group!" like the tests/integration/src/identity/user_group/add.rs - we should be using this pattern more since it includes automatic cleanup of the resources created by the test what helps to fix test reliability. It is actually already present in some of the added tests.

tests/integration/src/identity/group/create.rs
```diff
@@ -0,0 +1,52 @@
...
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+//! Test create group functionality.
+
+use eyre::Result;
+use tracing_test::traced_test;
+use uuid::Uuid;
+
+use openstack_keystone::identity::IdentityApi;
```
gtema: From now on this import is not necessary. Please remove it, but you actually should just run `cargo check -p test_integration --tests` to see warnings. Focus only in the files you add

gtema (changes requested):
generally good, but we need some refinements:
- add update tests as well
- try to add some negative tests (adding user to non-existing group or vice versa, using duplicate names, invalid data, delete not existing resource, etc)

gtema (changes requested):
since we now have DCO check which need to pass for all commits you are in unfortunate situation forcing you to either rewrite every single commit message , or, better, squash all changes to the single commit and put a proper commit message to it (you can do it with 'git rebase -i origin/main' - or it may be upstream/main). If you are not able to solve this (since this is a fork) you can follow a different approach and cherry-pick all 3 commits to the new branch (git cherry-pick -n 7241113a26c7e7363f5c904ecbc8ae947a20ebce, git cherry-pick -n .........). Another approach for not recreating the change is to do a `git reset --soft $(git merge-base HEAD main)` or just 3 times `git reset --soft HEAD^1`. This will undo commits while keeping the changes.

Which leaves the `git rebase -i origin/main` and pick the `reword` for every commit the easiest option.

Otherwise tests themselves looks fine

gtema (approved):

gtema (changes requested):
Ouch, there was an error in workflows causing the test not to be actually running at all. Please fix  errors and lonter complains

ShJ-code:
I reviewed the CI error and found that all errors are reporting the same type mismatch issue: the tests added within this PR call the `IdentityApi` provider methods and pass `&state` as their first argument. However, probably due to the upgrade of the API, every one of those methods takes an execution context as its first parameter. You might want to use the constructor `fn internal` from `ExecutionContext` (within `crates/core/src/auth.ts`) since the integration tests do not have authenticated caller. Specifically, the context could be built once in each test before passed in.

gtema:
> I reviewed the CI error and found that all errors are reporting the same type mismatch issue: the tests added within this PR call the `IdentityApi` provider methods and pass `&state` as their first argument. However, probably due to the upgrade of the API, every one of those methods takes an execution context as its first parameter. You might want to use the constructor `fn internal` from `ExecutionContext` (within `crates/core/src/auth.ts`) since the integration tests do not have authenticated caller. Specifically, the context could be built once in each test before passed in.

yes, there was a bigger restructure. Please adapt to how other tests initialize the ExecutionContext

gtema (changes requested):
Can you please address the merge conflict? We should be able to merge this change soon

gtema (approved):

PR #835, opened 2026-06-23
feat(auth): Password hashing parity with Python Keystone

magdang:
## Overview

This PR completes the password hashing implementation in the Rust Keystone
service so that hashes produced by the Rust service are verifiable by the
Python service and vice versa — enabling live database migration in either
direction without requiring a password reset.

### PasswordHasher trait

A `PasswordHasher` trait is introduced with `async fn hash` and
`async fn verify`. One zero-size struct per algorithm mirrors the class
layout in `keystone/common/password_hashers/`:

| Rust struct | Python class |
|-------------|-------------|
| `BcryptHasher` | `bcrypt.py::Bcrypt` |
| `BcryptSha256Hasher` | `bcrypt.py::Bcrypt_sha256` |
| `ScryptHasher` | `scrypt.py::Scrypt` |
| `Pbkdf2Sha512Hasher` | `pbkdf2.py::Sha512` |
| `PlaintextHasher` | Rust-only extension |

The public API (`hash_password`, `verify_password`,
`get_or_init_dummy_hash`, `generate_dummy_hash`) signatures are
unchanged — only the internal dispatch is restructured.

### Scrypt

Uses `scrypt::scrypt()` directly with Keystone's hardcoded parameters
(`ln=16, r=8, p=1, output=32 bytes`) and `STANDARD_NO_PAD` base64,
matching Python's `binascii.b2a_base64(...).rstrip(b"=\n")`. The PHC
wrapper (`Scrypt::hash_password`) is not used as it applies different
defaults.

### PBKDF2-SHA512

The `pbkdf2 = "0.12"` direct dependency is removed. PBKDF2-HMAC-SHA512
is implemented directly using the workspace `hmac v0.13` + `sha2 v0.11`
following RFC 2898 §5.2. This eliminates a `digest v0.10`/`v0.11` version
conflict that would otherwise arise from `pbkdf2 v0.12`'s internal sha2.

### Dummy-hash cache reset on config reload

`reset_dummy_hash_cache()` clears the `(algorithm, rounds)`-keyed dummy
hash cache. It is wired to `ConfigManager::notify_tx` in
`crates/keystone/src/bin/keystone.rs` via a `reset_dummy_hash_on_reload`
task that subscribes to the broadcast channel and clears the cache on
every config reload (lagged receivers are also treated as a reload since
clearing is always safe).

### Tooling

- `tools/generate_kats.py` — updated with the correct pre-dispatch
  truncation policy (`BCRYPT_MAX_LENGTH = 72`, `truncate()` helper)
  mirroring `verify_length_and_trunc_password`.
- `tools/cross_verify.py` — new script for bidirectional verification
  (see below).

## Verification suite

The PR ships a two-directional verification suite that guarantees
compatibility for anyone migrating between the Python and Rust services.

### How it works

**Direction 1 — Python produces, Rust verifies (KAT vectors)**

Seven known-answer test vectors are generated from the real, unmodified
Python Keystone hashers (`bcrypt 5.0`, `cryptography 49.0`) using
`tools/generate_kats.py`. These vectors are hardcoded into
`#[tokio::test]` cases and verified by `verify_password` on every CI
run — no Python runtime needed at test time.

Coverage: all 4 algorithms, including the 72-byte bcrypt truncation
boundary, 73-byte inputs to BcryptSha256/Scrypt/PBKDF2 (must NOT
truncate), and multi-byte UTF-8 passwords.

**Direction 2 — Rust produces, Python verifies (cross-verify)**

`tools/cross_verify.py` takes an algorithm name, a plaintext password,
and a hash string, imports the real `keystone.common.password_hashers`
classes, and exits 0 (verified), 1 (rejected), or 2 (error). It is
the definitive oracle for whether a Rust-produced hash is acceptable
to the Python service.

The four `test_cross_verify_*` tests in the Rust suite call this script
via `tokio::process::Command`. They are gated on the
`KEYSTONE_PYTHON_CHECKOUT` environment variable and skipped in standard
CI — set it to a real Keystone checkout to run them locally:

```bash
KEYSTONE_PYTHON_CHECKOUT=~/Projects/openstack/keystone \
  cargo test -p openstack-keystone-core -- cross_verify
```

Both directions were confirmed end-to-end before this PR was opened.
Correct password → exit 0 and wrong password → exit 1 for all four
algorithms in both directions.

### Running the verification suite

```bash
# Standard CI (KAT vectors only — no Python required)
cargo test -p openstack-keystone-core

# Full bidirectional (requires a real Keystone Python checkout)
KEYSTONE_PYTHON_CHECKOUT=~/Projects/openstack/keystone \
  cargo test -p openstack-keystone-core -- cross_verify

# Regenerate KAT vectors from Python (run from the Keystone checkout)
cd ~/Projects/openstack/keystone
python tools/generate_kats.py

# Spot-check a single Rust-produced hash against Python
python tools/cross_verify.py bcrypt "mypassword" "<hash>"
```

## Open items for discussion

- **Cache reset option** — implemented as option (a): pub-sub via
  `ConfigManager::notify_tx`. Option (b) (provider hook via
  `ProviderHooks`) is an alternative if preferred.
- **Missing config fields** — `scrypt_block_size`, `scrypt_parallelism`,
  `salt_bytesize`, `strict_password_check` are not yet in
  `IdentityProvider`; Keystone's hardcoded defaults are used. Suggested
  as follow-up issues.
- **ADR** — the `PasswordHasher` trait and cache lifecycle design may
  warrant an ADR per the spec process.
- **`$2$` / `$2x$` bcrypt idents** — recognised by `detect_algo` but
  not tested; low-priority follow-up.

Supersedes #793.

Note: This PR was prepared with the help of AI


crates/config/src/identity.rs
```diff
@@ -68,6 +68,12 @@ pub enum PasswordHashingAlgo {
     /// Bcrypt.
     #[default]
     Bcrypt,
+    /// Bcrypt combined with SHA256.
+    BcryptSha256,
+    /// Scrypt.
+    Scrypt,
+    /// PBKDF2 with SHA512.
+    Pbkdf2Sha512,
```
gtema, line 74: let's please sort them alphabetically

crates/core/src/common/password_hashing.rs
```diff
@@ -12,313 +12,1330 @@
...
-        #[from]
-        source: str::Utf8Error,
-    },
+// ---------------------------------------------------------------------------
+// PasswordHasher trait — mirrors the hash()/verify() static-method pair
+// every Python hasher class in keystone.common.password_hashers implements.
+// Dispatch is always static (no dyn), so async fn in traits is safe here.
+// ---------------------------------------------------------------------------
+
+trait PasswordHasher {
```
gtema: right, this is exactly what I was proposing.
Now you have actually a possibility to split this huge file into dedicated implementers per algorithm and keep unittests close and focused

crates/core/src/common/password_hashing.rs
```diff
@@ -12,313 +12,1330 @@
...
+        //   salt = salt_with_opts[-22:]
+        //   hmac_digest = base64.b64encode(hmac.digest(salt, password, "sha256"))
+        //   hashed = bcrypt.hashpw(hmac_digest, salt_with_opts)
+        //   digest = hashed[-31:]
+        //   return f"$bcrypt-sha256$v=2,t=2b,r={rounds}${salt}${digest}"
+        let password_bytes = password.to_vec();
+        let rounds = conf.identity.password_hash_rounds.unwrap_or(12);
+
+        let hash = task::spawn_blocking(move || {
+            // Generate the real salt ourselves instead of throwing away a
```
gtema: When reading this comment out of context of our previous discussion you see that it refers to older ideas, but here this makes no sense. Try to re-read all comments yourself without the older context - just read the code

crates/core/src/common/password_hashing.rs
```diff
@@ -12,313 +12,1330 @@
...
+            // alphabet characters) is a real trap: a bcrypt salt only
+            // carries 128 bits of entropy across 22 characters, so the
+            // last character has only 4 valid bits - only 4 of the 64
+            // alphabet characters are canonical there. A non-canonical
+            // salt gets silently re-canonicalized by bcrypt when the
+            // hash is computed, so the salt string you embed in the
+            // record stops matching the salt you used for the HMAC key.
+            // Letting the bcrypt crate generate it sidesteps this
+            // entirely.
+            let raw_salt: [u8; 16] = rand::random();
```
gtema: for the future I think we should move salt generation into a separate function. Sometimes there are requirements requiring to use special random generators (CSPRNG) and it would be easier to control when we have it centralized

crates/core/src/common/password_hashing.rs
```diff
@@ -12,313 +12,1330 @@
...
+    async fn verify(
+        &self,
+        _conf: &Config,
+        password: &[u8],
+        hash: &str,
+    ) -> Result<bool, PasswordHashError> {
+        // mirrors keystone/common/password_hashers/bcrypt.py::Bcrypt_sha256.verify()
+        // exactly: there is no "version" concept in the real implementation.
+        // It always HMACs (never falls back to a plain digest), and it
+        // does not even look at `v=` — it just scans every comma-delimited
```
gtema: this looks like a unicode char - we should not have non-ascii chars in our file with exceptions of those password templates and co

gtema (changes requested):
This looks great, thanks.

This is not a very thorough review, but gives a further recommendations to make the review easier (one huge file is hard to review)

crates/core/src/common/password_hashing.rs
magdang: I have divided password_hashing.rs into sub files for each module

crates/core/src/common/password_hashing.rs
magdang: Done in this update — salt generation is now centralized in `generate_salt()` in `password_hashing/mod.rs`. All four algorithms that need a salt (`BcryptSha256`, `Scrypt`, `Pbkdf2Sha512`, and `Bcrypt`) call it from there, so swapping the CSPRNG source is a one-line change.

crates/core/src/common/password_hashing/mod.rs
```diff
@@ -0,0 +1,615 @@
...
+        PasswordHashingAlgo::Pbkdf2Sha512 => {
+            Pbkdf2Sha512Hasher.verify(conf, truncated, hash_str).await
+        }
+        PasswordHashingAlgo::Scrypt => ScryptHasher.verify(conf, truncated, hash_str).await,
+        PasswordHashingAlgo::None => PlaintextHasher.verify(conf, truncated, hash_str).await,
+    }
+}
+
+/// Shared test helpers used by every per-algorithm submodule's `tests`.
+#[cfg(test)]
```
gtema: I understand the tension, but it should not be required to have 2 #[cfg(test)] submodules. Due to the fact that all hasher modules are child to this one they are able to access the method even without declaring it pub(crate). Merge this "mock_config" method into the general tests module

crates/core/src/common/password_hashing/mod.rs
```diff
@@ -0,0 +1,615 @@
...
+    //   KEYSTONE_PYTHON_CHECKOUT=~/Projects/openstack/keystone \
+    //     cargo test -p openstack-keystone-core -- cross_verify
+    //
+    // The checkout must have the `bcrypt` and `cryptography` packages installed.
+
+    /// Run a Rust-produced hash through tools/cross_verify.py against the Python
+    /// hashers. Returns the script's exit code (0 = verified, 1 = rejected,
+    /// 2 = error). Returns `None` when no Python checkout is configured so the
+    /// caller can skip.
+    async fn python_cross_verify(algo_name: &str, password: &str, hash: &str) -> Option<i32> {
```
gtema, line 534: this is a very neat approach. I need to admit I wasn't even thinking about it. It's great, but it has a problem (as comments suggest) that it requires python keystone to be installed for the test to work. It may start failing at any time and I wonder why it didn't fail already now. Let's keep it but also keep an eye on potential improving the CI setup to ensure python keystone libs are available

gtema (changes requested):
let's fix this one issue with the unnecessary test submodule and land the change. Afterwards we continue polishing it

crates/core/src/common/password_hashing/mod.rs
magdang: Done. Two issues were found and fixed:

1. cross_verify.py sys.path bug: When invoked via absolute path (as the Rust test does), Python sets sys.path[0] to the script's directory, not current_dir. So import keystone never actually resolved — the tests were only passing locally because they were silently skipping (no KEYSTONE_PYTHON_CHECKOUT set in CI). Fixed by inserting os.getcwd() into sys.path at the top of the script.
2. Incomplete deps: pip install bcrypt cryptography isn't enough. The import chain goes bcrypt.py → keystone.exception → keystone.conf → oslo_messaging, oslo_cache, ... — nearly the full Keystone stack. Fixed by cloning openstack/keystone in CI and running pip install -r requirements.txt.

The cross-verify tests now actually execute in CI (all 4 algorithms verified **locally** against the real Python hashers before pushing). They won't actually run through Github Actions until the PR is merged. So I can't really test it out yet.

crates/core/src/common/password_hashing/mod.rs
```diff
@@ -0,0 +1,613 @@
...
+fn mock_config(algo: openstack_keystone_config::PasswordHashingAlgo, max_len: usize) -> Config {
+    let mut conf = Config::default();
+    conf.identity.password_hashing_algorithm = algo;
+    conf.identity.password_hash_rounds = Some(12);
+    conf.identity.max_password_length = max_len;
+    conf
+}
+
+#[cfg(test)]
+mod tests {
```
gtema: please merge all of this into the single tests submodule

.github/workflows/ci.yml
```diff
@@ -57,7 +57,19 @@ jobs:
       - name: Rust Cache
         uses: swatinem/rust-cache@98c8021b550208e191a6a3145459bfc9fb29c4c0 # v2.8.0
 
+      - name: Set up Python for cross-verification tests
+        uses: actions/setup-python@v5
```
gtema: please use the tagged action for setup-python

.github/workflows/ci.yml
```diff
@@ -57,7 +57,19 @@ jobs:
...
+        with:
+          python-version: '3.x'
+
+      - name: Clone Python Keystone and install dependencies
```
gtema: it's actually not necessary to clone the repo. What you need to ensure is that keystone is installed for the python that is invoked by the test. It should be enough just to do "pip install keystone" and that's it, not even to set the checkout env

gtema (changes requested):

gtema (approved):

PR #837, opened 2026-06-24
Implement password complexity enforcement

### LINKED ISSUE #401: PCI-DSS 8.2.3: Password complexity (min 7 chars, numeric + alphabetic)
Implement and document

MohidSheraz:
Closes #401

This completes the SQL identity enforcement path for PCI-DSS 8.2.3 password complexity.

The config layer already supported `password_regex`, `password_regex_description`, regex compilation, and `validate_password()`. This change calls password validation before password history checks, hashing, expiry calculation, and password storage.

Tested with:

```bash
cargo test -p openstack-keystone-identity-driver-sql test_set_new_password_rejects_password_regex_mismatch
cargo test -p openstack-keystone-identity-driver-sql password

### DIFF
--- a/crates/identity-driver-sql/src/password.rs
+++ b/crates/identity-driver-sql/src/password.rs
@@ -57,6 +57,9 @@ pub async fn set_new_password<C: ConnectionTrait>(
     password: SecretString,
     existing_passwords: Vec<db_password::Model>,
 ) -> Result<db_password::Model, IdentityProviderError> {
+    conf.security_compliance
+        .validate_password(&password)
+        .map_err(IdentityProviderError::SecurityCompliance)?;
     check_history::check_password_history(conf, &existing_passwords, &password).await?;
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: crates/identity-driver-sql/src/password.rs [hunk 1]
```diff
@@ -283,4 +286,31 @@ pub mod tests {
             "password is expired"
         );
     }
+    #[tokio::test]
```
gtema, line 289: please add an empty line above to keep it separated from previous test

gtema (changes requested):
you should always take care of synchronizing you fork repository with the origin. When you start working on the issue and create a branch ensure it is branched from the main branch and not from another branch (once you pushed the PR it again contained another commit for the test you have added in another PR). When you create PR or push a new revision do yourself a sanity check just looking at the commits tab whether everything is right

Issue #734, opened 2026-05-27
Implement default redirects for trailing slash in the url


gtema:
It is possible in the Axum to add a [generic processing](https://docs.rs/tower-http/latest/tower_http/normalize_path/index.html) that will allow to serve url with and without trailing slash to be served by one handler. Python keystone is not sensitive to this and so the rust keystone should also work good with it

PR #842, opened 2026-06-24
feat(api): capture client connection info via ConnectInfo

### LINKED ISSUE #358: Start capturing client information in the request processing
In order to implement IP based control of the login we need to start capturing client information (IP address and so on). https://docs.rs/axum/latest/axum/struct.Router.html#method.into_make_service_with_connect_info may describe how this can be achieved.

magdang:
Closes #358.

## What this does

Starts capturing client connection information (the client IP) in request processing, as the enabling step for a future IP-based login-control feature. The lockout/rate-limit mechanism itself is **out of scope** here — this is the plumbing that makes the client address available to handlers and middleware.

- Switches the **public** listener's make-service to `into_make_service_with_connect_info::<SocketAddr>()`, so the raw TCP peer address is stored in a `ConnectInfo<SocketAddr>` request extension.
- Surfaces it on the existing `request` tracing span as a `client.addr` field (reusing the previously-empty placeholder slot), so the capture is verifiable in request logs.

## Mirrors Python Keystone

This deliberately follows how upstream Python Keystone captures the client address:

| Python | This PR |
|---|---|
| WSGI `REMOTE_ADDR` / `flask.request.remote_addr` (raw peer when no proxy parsing applies) | `ConnectInfo<SocketAddr>` |
| `remote_addr` surfaced in the auth-failure log line | `client.addr` field on the `request` span |
| Proxy-header parsing kept separate, **config-gated** via `[oslo_middleware] enable_proxy_headers_parsing` (default off) | deliberately deferred — see follow-up below |

Python captures a single `REMOTE_ADDR` at the WSGI front door; the analogue here is the public HTTP interface. The internal (SPIFFE mTLS/TCP) and admin (SPIFFE mTLS/UDS) interfaces are intentionally **not** covered: they bypass `axum::serve`, and a Unix socket has no meaningful `SocketAddr`.

## Important: raw peer only — not proxy-aware

The captured address is the **raw TCP peer**. Behind a reverse proxy / load balancer (the normal production topology) it is the proxy's address, not the real client. It must **not** be used directly for IP-based login control until a trusted forwarded-header layer exists.

## Required follow-up (after this merges)

A config-gated forwarded-header layer that parses `X-Forwarded-For` / RFC 7239 `Forwarded`, mirroring Python's `[oslo_middleware] enable_proxy_headers_parsing` — **off by default**, so a deployment that is not actually behind a trusted proxy cannot be tricked into trusting a spoofed header. (`axum-client-ip` is a candidate, or a hand-rolled middleware.) This is required before any IP-based login control is built on top of this capture.

## Composition with #734

Issue #734 wraps the router with `NormalizePathLayer` from the outside, so the public listener serves a `NormalizePath<Router>`, not a plain `Router`. `into_make_service_with_connect_info` is available on the same `axum::ServiceExt` trait already in use (blanket impl for any `Service`), so it applies directly — no restructuring needed.

## Tests

Added `connect_info_is_captured_and_normalizes`, which drives the exact production make-service path fully in-process and asserts a `/echo/` request **both** normalizes the trailing slash **and** delivers the injected peer address to the handler — proving #358 and #734 compose. The existing `trailing_slash_is_normalized` regression still passes.

### DIFF
--- a/crates/keystone/src/bin/keystone.rs
+++ b/crates/keystone/src/bin/keystone.rs
@@ -16,13 +16,14 @@
 //! This is the entry point of the `keystone` binary.
 
 use std::io;
+use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
 
 use axum::{
     Router, ServiceExt,
-    extract::DefaultBodyLimit,
+    extract::{ConnectInfo, DefaultBodyLimit},
     http::{self, HeaderName, Request, header},
 };
 use clap::{Parser, ValueEnum};
@@ -348,10 +349,19 @@ async fn main() -> Result<(), Report> {
         .layer(
             TraceLayer::new(common::KeystoneResponseClassifier)
                 .make_span_with(|request: &Request<_>| {
+                    // Raw TCP peer address captured by
+                    // `into_make_service_with_connect_info` on the public
+                    // listener (the keystone-ng analogue of Python Keystone's
+                    // WSGI REMOTE_ADDR / flask.request.remote_addr). `None` for
+                    // the SPIFFE interfaces, which do not populate ConnectInfo.
+                    let client_addr = request
+                        .extensions()
+                        .get::<ConnectInfo<SocketAddr>>()
+                        .map(|ConnectInfo(addr)| *addr);
                     info_span!(
                         "request",
                         method = ?request.method(),
-                        some_other_field = tracing::field::Empty,
+                        client.addr = ?client_addr,
                         uri = ?request.uri().path(),
                         x_request_id = ?request.headers().get("x-openstack-request-id")
                     )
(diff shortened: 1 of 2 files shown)


crates/keystone/src/bin/keystone.rs
```diff
@@ -486,12 +496,25 @@ async fn main() -> Result<(), Report> {
...
+                // `rest_app` is `NormalizePath<Router>` (issue #734 wraps the
+                // Router from the outside), which has no
+                // `Router::into_make_service_with_connect_info`; use axum's
+                // `ServiceExt` (blanket-impl'd for any `Service`) with an
                 // explicit request type to satisfy inference (E0284).
+                //
+                // `into_make_service_with_connect_info::<SocketAddr>` stores the
+                // raw TCP peer address in a `ConnectInfo<SocketAddr>` request
+                // extension (the analogue of Python Keystone's WSGI REMOTE_ADDR).
+                // This is the *raw* peer, not proxy-resolved: behind a reverse
```
gtema, line 508: This need to be addressed of course as well. Follow-up should be absolutely fine

crates/keystone/src/bin/keystone.rs
```diff
@@ -348,10 +349,19 @@ async fn main() -> Result<(), Report> {
         .layer(
             TraceLayer::new(common::KeystoneResponseClassifier)
                 .make_span_with(|request: &Request<_>| {
+                    // Raw TCP peer address captured by
+                    // `into_make_service_with_connect_info` on the public
+                    // listener (the keystone-ng analogue of Python Keystone's
+                    // WSGI REMOTE_ADDR / flask.request.remote_addr). `None` for
+                    // the SPIFFE interfaces, which do not populate ConnectInfo.
```
gtema, line 356: this is an interesting point. I agree for spiffe we currently go absolutely different way, but it should be possible to capture the information there as well. You can definitely capture the remote IP address and at a later phase the headers (for x-forward) should be also available. Please evaluate this in the followup

gtema (approved):
Nice, but would need a followup

Issue #843, opened 2026-06-25
Implement rate limiting


gtema:
We have ADR-0022-rate-limiting which describes how rate limiting should be implemented. There is a slight dependency on #358 which should introduce client data that is a base for limiting usually

PR #846, opened 2026-06-25
feat(api): Add global IP rate limiting framework (ADR-0022 phase 1)

magdang:
## Summary

- Implements ADR-0022 phase 1: a handler-level, `governor`-backed rate-limiting framework wired on `POST /v3/auth/tokens`.
- Adds a reusable `RateLimitSection` config struct and `RateLimitState` service field so future buckets (per-user, per-domain, per-IdP) are a one-field addition to the framework.
- Fires the global per-IP check **before** password hashing (ADR-0022 Invariant 4); returns `429 Too Many Requests` with `Retry-After`.

> **#842 (ConnectInfo capture) is now merged into `main`**, so this PR targets `main` directly and is no longer stacked. Rebased onto latest `main`; conflicts with the concurrently-merged API Key limiter (ADR-0021) and audit refactor are resolved (see "Integration notes" below).

## ADR-0022 invariants addressed

| # | Invariant | How |
|---|-----------|-----|
| 1 | No hardcoded limits | All limits from `[rate_limit_global_ip]` in `keystone.conf` |
| 2 | Fail-hard init | `RateLimitState::from_config` returns `KeystoneError::RateLimitConfig` and aborts startup when `enabled=true` with zero burst or replenish rate |
| 3 | Uniform response | Single `Retry-After` header; no key-identifying information exposed |
| 4 | Check before hash | Rate limit fires before `authenticate_request` (password hash) |
| 5 | Distinct buckets | One `Arc<DefaultKeyedRateLimiter<String>>` per bucket; deferred buckets are `None` fields |
| 6 | Monotonic clock | `governor::DefaultClock` = `QuantaClock` on std targets (TSC-backed, always monotonic) |

Invariants 7 (username normalization) and 8 (post-lookup per-user throttle) are deferred — they require keying on a confirmed user ID after DB lookup but before hashing, tracked as a follow-up driver refactor.

## SPIFFE bypass

Internal (mTLS/TCP) and admin (mTLS/UDS) interfaces do not populate `ConnectInfo<SocketAddr>`; the handler receives `None` and skips IP limiting. Only the public TCP listener is subject to this check.

Note: `Option<Extension<ConnectInfo<SocketAddr>>>` is used (not `Option<ConnectInfo<SocketAddr>>`) because in axum 0.8, `Option<T>: FromRequestParts<S>` requires `T: OptionalFromRequestParts<S>`, which `ConnectInfo<T>` does not implement but `Extension<T>` does.

## Integration notes (post-rebase)

`main` gained an **API Key ingress limiter** (`api_key_rate_limiter`, ADR-0021) and an **audit refactor** that splits `create` into an outer audit wrapper + inner `create_inner`. This PR integrates with both:

- Both limiters coexist as independent fields on `Service` (`api_key_rate_limiter` + `rate_limiters`).
- The global-IP check lives in `create_inner`, so a rate-limited request still flows through the outer handler's perimeter-authenticate **audit** emission (recorded as a `TooManyRequests` failure).
- **`KeystoneApiError::TooManyRequests` consolidated:** `main`'s API Key limiter returned a bare unit `TooManyRequests` (429, no `Retry-After`). I merged it into the ADR-0022 struct variant `TooManyRequests { retry_after }`, and updated the API Key path to compute a real `retry_after` from its limiter. Both paths now emit a uniform 429 body + `Retry-After` header (ADR-0022 Invariant 3).

## ❓ Question for maintainer

I **unified** the two `TooManyRequests` variants (ADR-0021 API Key limiter + ADR-0022 IP limiter) into one `{ retry_after }` variant so every 429 carries a `Retry-After`. This slightly changes the API Key limiter's response (it now includes `Retry-After`, which it didn't before). **Is that the desired behavior, or would you prefer the two limiters keep separate error variants / response shapes?** Easy to split back out if you want them decoupled.

## Files changed

| File | Change |
|------|--------|
| `crates/config/src/rate_limit.rs` *(new)* | `RateLimitSection` — reusable config struct for any bucket |
| `crates/config/src/lib.rs` | Add `rate_limit_global_ip: RateLimitSection` to `Config` |
| `crates/core-types/src/error.rs` | Add `KeystoneError::RateLimitConfig` |
| `crates/core/src/rate_limit.rs` *(new)* | `RateLimitState`, `check_ip`, `retain_recent`, IPv6 /64 key derivation, unit tests |
| `crates/core/src/keystone.rs` | Add `rate_limiters: RateLimitState` to `Service` |
| `crates/core/src/api/api_key_auth.rs` | API Key limiter uses the unified `TooManyRequests { retry_after }` |
| `crates/api-types/src/error.rs` | Consolidate `TooManyRequests { retry_after }` variant |
| `crates/api-types/src/error_conv.rs` | Map `TooManyRequests` → 429 + `Retry-After` header |
| `crates/keystone/src/api/v3/auth/token/create.rs` | Add IP rate-limit check to `create_inner`; handler + e2e tests |
| `crates/keystone/src/audit.rs` | Match updated `TooManyRequests { .. }` variant |
| `crates/keystone/src/bin/keystone.rs` | Add 60 s background eviction task |
| `tools/keystone.conf` | Document `[rate_limit_global_ip]` with default values |

## Test plan

- [x] `cargo fmt --check` clean
- [x] `cargo clippy --lib --tests --workspace` clean
- [x] Unit tests pass across `core` / `config` / `api-types` / `keystone` (759 total, incl. new rate-limit + API Key limiter tests)
- [x] `cargo build --locked` succeeds — no `Cargo.lock` changes needed (`governor` was already a dependency via ADR-0021)
- [x] **End-to-end `401 → 429` flip verified by an automated test** (`test_rate_limit_429_over_connect_info_make_service`): drives the real `create` route through the exact `into_make_service_with_connect_info::<SocketAddr>` path the public listener uses, with a fixed peer address so `ConnectInfo` is populated by axum itself (not injected). First request from an IP is non-429; the second (burst spent) returns `429` **with** a `Retry-After` header — the same behavior the manual `curl` loop would show, without needing a live DB/OPA.

Partially implements #843.

### DIFF
--- a/crates/api-types/src/error.rs
+++ b/crates/api-types/src/error.rs
@@ -76,6 +76,17 @@ pub enum KeystoneApiError {
     #[error("selected authentication is forbidden")]
     SelectedAuthenticationForbidden,
 
+    /// Rate limit exceeded (HTTP 429 Too Many Requests).
+    ///
+    /// `retry_after` is the number of seconds the client must wait before
+    /// retrying. Surfaced via the `Retry-After` response header (ADR-0022,
+    /// Invariant 3). No key-identifying information is included.
+    #[error("Rate limit exceeded. Retry in {retry_after}s.")]
+    TooManyRequests {
+        /// Minimum seconds to wait before retrying.
+        retry_after: u64,
+    },
+
     /// (de)serialization error.
     #[error(transparent)]
     Serde {
@@ -86,16 +97,12 @@ pub enum KeystoneApiError {
     #[error("missing x-subject-token header")]
     SubjectTokenMissing,
 
-    #[error("rate limit exceeded, retry later")]
-    TooManyRequests,
-
     /// Request is syntactically valid but semantically invalid (RFC 4918
     /// §11.2 / RFC 7231 §6.5.1's `422` successor). Used for write-time
     /// validation failures that are not simple malformed-request errors
     /// (e.g. ADR 0024 §2.C's `Authorization::Project` prohibition).
     #[error("{0}.")]
     UnprocessableEntity(String),
-
     #[error("The request you have made requires authentication.")]
     UnauthorizedNoContext,
 
--- a/crates/api-types/src/error.rs
+++ b/crates/api-types/src/error.rs
@@ -81,7 +81,7 @@ pub enum KeystoneApiError {
     /// `retry_after` is the number of seconds the client must wait before
     /// retrying. Surfaced via the `Retry-After` response header (ADR-0022,
     /// Invariant 3). No key-identifying information is included.
-    #[error("Rate limit exceeded. Retry in {retry_after}s.")]
+    #[error("Rate limit exceeded. Retry in {retry_after} seconds.")]
     TooManyRequests {
         /// Minimum seconds to wait before retrying.
         retry_after: u64,
--- a/crates/api-types/src/error_conv.rs
+++ b/crates/api-types/src/error_conv.rs
@@ -52,8 +52,7 @@ impl IntoResponse for KeystoneApiError {
                     "message": self.to_string(),
                 }
             }));
-            let retry_value = HeaderValue::from_str(&retry_after.to_string())
-                .unwrap_or_else(|_| HeaderValue::from_static("60"));
+            let retry_value = HeaderValue::from(*retry_after);
             let mut response = (StatusCode::TOO_MANY_REQUESTS, body).into_response();
             response
                 .headers_mut()
@@ -674,6 +673,7 @@ mod tests {
     #[test]
     fn too_many_requests_returns_429_with_retry_after() {
         let err = KeystoneApiError::TooManyRequests { retry_after: 42 };
+        assert_eq!(err.to_string(), "Rate limit exceeded. Retry in 42 seconds.");
         let response = <KeystoneApiError as IntoResponse>::into_response(err);
         assert_eq!(response.status(), StatusCode::TOO_MANY_REQUESTS);
         let retry_after = response
@@ -681,18 +681,5 @@ mod tests {
             .get(header::RETRY_AFTER)
             .expect("Retry-After header must be present");
         assert_eq!(retry_after.to_str().unwrap(), "42");
-    }
-
-    #[test]
-    fn too_many_requests_retry_after_fallback_on_large_value() {
-        // u64::MAX cannot fit in a header value — the fallback must be "60".
-        // In practice our handler always passes a small Duration::as_secs()
-        // value, but the fallback path must be covered.
-        // We verify the fallback by constructing a HeaderValue that fails.
-        let bad_value = "not\na valid\nheader";
-        let result = HeaderValue::from_str(bad_value);
-        assert!(result.is_err(), "sanity: newlines must be rejected");
-        let fallback = result.unwrap_or_else(|_| HeaderValue::from_static("60"));
-        assert_eq!(fallback.to_str().unwrap(), "60");
     }
 }
--- a/crates/api-types/src/error_conv.rs
+++ b/crates/api-types/src/error_conv.rs
@@ -16,7 +16,7 @@ use {
     axum::{
         Json,
         extract::rejection::JsonRejection,
-        http::StatusCode,
+        http::{HeaderValue, StatusCode, header},
         response::{IntoResponse, Response},
     },
     serde_json::json,
@@ -42,6 +42,25 @@ use crate::error::KeystoneApiError;
 
 impl IntoResponse for KeystoneApiError {
     fn into_response(self) -> Response {
+        // Rate-limit rejections need a `Retry-After` header in addition to the
+        // JSON body, so they are handled before the generic status-code path
+        // (ADR-0022 Invariants 3 and 4).
+        if let KeystoneApiError::TooManyRequests { retry_after } = &self {
+            let body = Json(json!({
+                "error": {
+                    "code": StatusCode::TOO_MANY_REQUESTS.as_u16(),
+                    "message": self.to_string(),
+                }
+            }));
+            let retry_value = HeaderValue::from_str(&retry_after.to_string())
+                .unwrap_or_else(|_| HeaderValue::from_static("60"));
+            let mut response = (StatusCode::TOO_MANY_REQUESTS, body).into_response();
+            response
+                .headers_mut()
+                .insert(header::RETRY_AFTER, retry_value);
+            return response;
+        }
+
         let status_code = match self {
             KeystoneApiError::Conflict(_) => StatusCode::CONFLICT,
             KeystoneApiError::NotFound { .. } => StatusCode::NOT_FOUND,
@@ -55,7 +74,6 @@ impl IntoResponse for KeystoneApiError {
             KeystoneApiError::InternalError(_) | KeystoneApiError::Other(..) => {
                 StatusCode::INTERNAL_SERVER_ERROR
             }
-            KeystoneApiError::TooManyRequests => StatusCode::TOO_MANY_REQUESTS,
             KeystoneApiError::UnprocessableEntity(_) => StatusCode::UNPROCESSABLE_ENTITY,
             _ => StatusCode::BAD_REQUEST,
         };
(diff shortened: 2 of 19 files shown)


crates/config/src/rate_limit.rs
```diff
@@ -0,0 +1,122 @@
...
+#[derive(Debug, Deserialize, Clone)]
+pub struct RateLimitSection {
+    /// When `false` (the default) the corresponding `governor` limiter is not
+    /// instantiated and the handler bypasses this check entirely.
+    #[serde(default)]
+    pub enabled: bool,
+
+    /// Maximum number of cells that can be consumed in a burst before
+    /// replenishment kicks in. Must be ≥ 1 when `enabled = true`.
+    #[serde(default = "default_burst_size")]
```
gtema, line 68: we should add validation for both values to be [1, 100000] - this is defined in the ADR

gtema (changes requested):
looks good, few nits inline
One more thing is that you implement here the IP address based throttling, but the invariant 9 is not addressed, which claims that in case of reverse proxies it should be the original address and not the proxy address. On one side this correlates with your other PR, on the other side this requires [rate_limit_trusted_proxies] section. It can be implemented after, but maybe good to cover now with introduction of the IP based limiter (don't know how phase1 was defined)

crates/config/src/rate_limit.rs
magdang: Done — added the `[1, 100000]` bound for both `burst_size` and `replenish_rate_per_second` (ADR-0022 config-bounds table). Enforced in `build_limiter`/`from_config` as a fail-hard startup error alongside the existing zero check, via a small `validated_scalar` helper. I kept it there rather than a field-level `validator` `range(min=1, max=100000)` because a *disabled* section must be allowed to carry out-of-range/zero values without aborting startup (existing invariant + tests) — a field-level range would fire unconditionally. Added tests for the upper bound and the 100000 boundary; field docs updated.

magdang:
On the two points:

**Config bounds (inline):** done — `burst_size` and `replenish_rate_per_second` are now enforced to `[1, 100000]` as a fail-hard startup error.

**Invariant 9 (trusted-proxy source IP):** you're right that with the direct-peer address the limiter buckets by the reverse proxy when one is present. I'd like to land it as a focused follow-up rather than in this phase-1 PR, for a concrete reason: ADR-0022 Invariant 7 requires the client-IP resolution to be *"a single shared utility used by both the rate limiter and the authentication pipeline"*. That shared utility is exactly what I just introduced in the sibling PR #908 — `core::api::forwarded::resolve_client_ip` (trusted-proxy allowlist + rightmost-non-trusted-hop walk + hop cap, already used by the API-key ingress). 

Implementing Invariant 9 here now would mean duplicating that resolver on this branch (the two PRs are on independent branches), which is the opposite of the "single shared utility" requirement. Once #908 lands, the follow-up is small: add a `trusted_proxies` source (either a `[rate_limit_trusted_proxies]` section as the ADR sketches, or reuse `[oslo_middleware] trusted_proxies`), resolve the client IP in the token-create extractor before `check_ip`, and warn when the global-IP limiter is enabled with an empty allowlist (ADR-0022 §Consequences). Happy to reorder if you'd prefer it in this PR instead — just flagging the duplication trade-off.

crates/core/src/rate_limit.rs
```diff
@@ -0,0 +1,402 @@
...
+
+impl RateLimitState {
+    /// Build [`RateLimitState`] from the application configuration.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`KeystoneError::RateLimitConfig`] and aborts startup when a
+    /// bucket is `enabled = true` but has a `burst_size` or
+    /// `replenish_rate_per_second` of zero (ADR-0022 Invariant 2).
+    pub fn from_config(cfg: &Config) -> Result<Self, KeystoneError> {
```
gtema: what if we implement auto-reloading from config similarly to the password hashing - register the hook for ConfigManager to inform RateLimiting of the new config

crates/keystone/src/api/v3/auth/token/create.rs
```diff
@@ -75,10 +88,23 @@ pub(super) async fn create(
...
     req.validate()?;
+
+    // Global per-IP rate-limit check (ADR-0022, Invariant 4).
+    // Fires BEFORE authenticate_request to avoid consuming CPU on password
+    // hashing for rejected requests.
+    if let Some(Extension(ConnectInfo(addr))) = client_addr
+        && let Err(retry_after) = state.rate_limiters.check_ip(addr.ip())
+    {
+        return Err(KeystoneApiError::TooManyRequests {
+            retry_after: retry_after.as_secs().max(1),
```
gtema: `check_ip` already applies `max(1)`

crates/core/src/rate_limit.rs
```diff
@@ -0,0 +1,402 @@
...
+    /// exceeded its quota.
+    ///
+    /// SPIFFE interfaces (internal / admin) do not populate
+    /// `ConnectInfo<SocketAddr>`, so the caller holds an
+    /// `Option<ConnectInfo<…>>` and skips this call when it is `None`.
+    pub fn check_ip(&self, ip: IpAddr) -> Result<(), Duration> {
+        let Some(ref limiter) = self.global_ip_limiter else {
+            return Ok(()); // bucket disabled — bypass
+        };
+        let key = rate_limit_key_for_ip(ip);
```
gtema: Here we allocate a new string on every request - this is sub-optimal. Keying on u32 (IPv4) and [u8; 8]-backed newtype (IPv6 /64 prefix) would eliminate per-request allocation entirely. The DefaultKeyedRateLimiter uses a HashMap, so any type implementing Hash + Eq works.

crates/api-types/src/error_conv.rs
```diff
@@ -588,4 +606,29 @@ mod tests {
...
+        assert_eq!(response.status(), StatusCode::TOO_MANY_REQUESTS);
+        let retry_after = response
+            .headers()
+            .get(header::RETRY_AFTER)
+            .expect("Retry-After header must be present");
+        assert_eq!(retry_after.to_str().unwrap(), "42");
+    }
+
+    #[test]
+    fn too_many_requests_retry_after_fallback_on_large_value() {
```
gtema: test name is not suitable, you are not testing rate limiting at all. Instead the test only verifies header construction

crates/api-types/src/error.rs
```diff
@@ -76,6 +76,17 @@ pub enum KeystoneApiError {
     #[error("selected authentication is forbidden")]
     SelectedAuthenticationForbidden,
 
+    /// Rate limit exceeded (HTTP 429 Too Many Requests).
+    ///
+    /// `retry_after` is the number of seconds the client must wait before
+    /// retrying. Surfaced via the `Retry-After` response header (ADR-0022,
+    /// Invariant 3). No key-identifying information is included.
+    #[error("Rate limit exceeded. Retry in {retry_after}s.")]
```
gtema: let's do "Rate limit exceeded. Retry in {retry_after} seconds." instead

gtema (changes requested):
Some efficiency optimizations are necessary

crates/core/src/rate_limit.rs
magdang: Done. `RateLimitState` now uses an `ArcSwap` snapshot and supports runtime `reload`. The Keystone binary subscribes to `ConfigManager` notifications and swaps the limiter state only when the relevant rate-limit/trusted-proxy config changes. Invalid runtime reloads keep the last-known-good limiter and log an error, while startup validation remains fail-hard. Added tests for unchanged, changed, disabled, enabled, and invalid reload behavior.

crates/keystone/src/api/v3/auth/token/create.rs
magdang: Done. Removed the handler-side `.max(1)` and now uses the retry duration returned by `check_ip`, since `check_ip` already guarantees the minimum retry-after value.

crates/core/src/rate_limit.rs
magdang: Done. IP limiter keys are now allocation-free: IPv4 buckets use `u32`, and IPv6 buckets use the first 64 bits as `[u8; 8]` inside a hashable enum. Added tests covering IPv4 keying, IPv6 `/64` keying, and shared IPv6 privacy-address quota behavior.

crates/api-types/src/error_conv.rs
magdang: Done. Removed the misleading fallback/header-construction test and updated the 429 response test to assert the actual `TooManyRequests { retry_after }` response contract: status, exact message, and `Retry-After`.

crates/api-types/src/error.rs
magdang: Done. The 429 message is now exactly: `Rate limit exceeded. Retry in {retry_after} seconds.`

magdang:
Trusted-proxy / ADR-0022 invariant 9 follow-up is now implemented in this PR as well. Added `[rate_limit_trusted_proxies]`, parse/validate trusted proxy CIDRs, and use the shared rightmost-non-trusted-hop resolver before selecting the IP rate-limit bucket. `check_ip` now accepts the forwarded-header value plus optional TCP peer and resolves the effective client internally. Also added the warning when IP limiting is enabled without trusted proxies, while preserving SPIFFE/admin requests without peer info as exempt.

gtema (approved):
Ok, few nits to be addressed in the next phases (please consider):

crates/core/src/rate_limit.rs:L646-759 (keyed limiter growth) risk: DefaultKeyedRateLimiter HashMap grows one entry per unique key between retain_recent() ticks (60s, keystone.rs:L1484). A real (non-spoofed, TCP-handshake-completing) botnet with many source IPs floods memory for up to 60s before eviction. Acceptable per ADR §Consequences, but consider capping store size or shortening the eviction interval if under active abuse — worth a one-line comment on the tradeoff since it's not obvious from rate_limit.rs alone.

crates/core/src/rate_limit.rs:L744-746 (check_ip retry floor) nit: wait.max(Duration::from_secs(1)) always floors Retry-After to 1s even when actual replenish is faster (e.g. replenish_rate_per_second=100000). Harmless (conservative), but worth a one-line comment — reader will wonder why real quota math is discarded.

magdang:
Thanks for the feedback. It would be really helpful to create an issue regarding this so it doesn't get forgotten. Also, please assign it to me.

PR #904, opened 2026-07-02
test(auth): Cover token rescope to newly granted project

### LINKED ISSUE #763: Improve API test test_rescope_project_scope
A test in https://github.com/openstack-experimental/keystone/blob/main/tests/api/src/auth/token/token.rs#L25 tries to re-authenticate to a project the user has access to according to the list of accessible projects. What the test misses is actually to create new project, grant access to it and verify user can rescope to it.
Right now such test would fail since https://github.com/openstack-experimental/keystone/blob/main/crates/keystone/src/api/v3/auth/token/create.rs#L61 falsely assumes token is bound to the project and rescoping is not allowed. This is not the case when authenticating using regular tokens. This check should be most likely removed completely since token_restrictions are controlling whether rescope is allowed or not and issue_token_context would validate whether the user has access to it.

ShJ-code:
Update the v3 token rescope API test to create a fresh user and two projects, grant the user a role on both projects, and verify that the user can rescope from one project to the other using a regular token.

This covers the missing case from #763: the target project is explicitly created and granted to the user before rescope is attempted, instead of only re-authenticating against projects already listed as accessible.

The test verifies both project-id and project-name/domain rescope paths, and uses `AsyncOpenStack::authorize` instead of `TestClient`.

Add a small `test_api::assignment::add_project_grant` helper for granting a project role to a user.

Closes: #763

### DIFF
--- a/tests/api/tests/api_v3/auth/token/token.rs
+++ b/tests/api/tests/api_v3/auth/token/token.rs
@@ -12,54 +12,134 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-use std::sync::Arc;
-
-use eyre::Result;
+use std::{collections::HashMap, sync::Arc};
 
 //use openstack_keystone_api_types::scope::*;
-use openstack_sdk::{AsyncOpenStack, config::CloudConfig};
+use eyre::Result;
+use openstack_keystone_api_types::v3::{project::ProjectCreateBuilder, user::UserCreateBuilder};
+use openstack_sdk::{
+    AsyncOpenStack,
+    auth::authtoken::AuthTokenScope,
+    config::CloudConfig,
+    types::identity::v3::{Domain as SdkDomain, Project as SdkProject},
+};
+use uuid::Uuid;
 
+use test_api::assignment::grant::add_project_grant;
 use test_api::auth::project::list_auth_projects;
+use test_api::common::get_session_by_user_password;
+use test_api::guard::ResourceGuard;
+use test_api::identity::user::create_user;
+use test_api::resource::project::create_project;
+use test_api::role::list_roles;
 
 #[tokio::test]
 async fn test_rescope_project_scope() -> Result<()> {
-    let mut test_client = AsyncOpenStack::new(&CloudConfig::from_env()?).await?;
+    let admin = Arc::new(AsyncOpenStack::new(&CloudConfig::from_env()?).await?);
+    let password = "TestPassword123!";
 
-    let projects = list_auth_projects(&Arc::new(test_client.clone())).await?;
+    let user = create_user(
+        &admin,
+        UserCreateBuilder::default()
+            .name(format!("usr_{}", Uuid::new_v4().simple()))
+            .domain_id("default")
+            .enabled(true)
+            .password(password)
+            .build()?,
+    )
+    .await?;
 
-    for project in projects {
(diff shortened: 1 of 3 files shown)


tests/api/tests/api_v3/auth/token/token.rs
```diff
@@ -12,54 +12,138 @@
...
-                false,
-                false,
-            )
-            .await?;
-    }
+    let user_sdk = get_session_by_user_password(&user.name, &user.domain_id, password).await?;
+    let projects = list_auth_projects(&user_sdk).await?;
+    assert!(projects.iter().any(|p| p.id == target.id));
 
+    let mut client = TestClient::default()?;
```
gtema: please do not use `TestClient` - we need to get rid of it. Instead you can use https://docs.rs/openstack_sdk/latest/openstack_sdk/struct.AsyncOpenStack.html#method.authorize what your `user_sdk` currently is.

crates/keystone/src/api/v3/auth/token/create.rs
```diff
@@ -84,14 +84,12 @@ async fn create_inner(
     let provider_scope: Option<ProviderScope> = req.auth.scope.clone().map(Into::into);
     let authz_info = get_authz_info(state, provider_scope.as_ref()).await?;
 
-    // This is a new authentication/reauthentication. Check if that is allowed at
-    // all
-    if let Some(token_restriction) = ctx.token_restriction()
```
gtema, line 132: I am very unsure in that . The lines that the issue were referring to (iirc) were removed with https://github.com/openstack-experimental/keystone/commit/466a3989c39c6d37e3056f3e05806ebdb7b8349d so you should not be required to change anything here

crates/core-types/src/auth.rs
```diff
@@ -679,11 +679,30 @@ impl SecurityContext {
                 }
             }
             ScopeInfo::Project { project, .. } => {
-                if let Some(token_restriction) = &self.token_restriction
```
gtema: this change is unnecessary (at least within this change). The test you modify is not dealing with restricted token, it only uses regular authentication tokens. Moreover I plan on removing token restrictions as a concept completely since they are now superseded by a more universal solution.

tests/api/tests/api_v3/auth/token/token.rs
```diff
@@ -12,54 +12,138 @@
...
+        UserCreateBuilder::default()
+            .name(format!("usr_{}", Uuid::new_v4().simple()))
+            .domain_id("default")
+            .enabled(true)
+            .password(password)
+            .build()?,
+    )
+    .await?;
+
+    let source = create_project(
```
gtema: those are not source and target, but rather project_a, project_b

gtema (changes requested):

ShJ-code:
Thanks for the comments. I will investigate those issues as well as CI checks and get back with changes ASAP.

ShJ-code:
I updated the PR to address the review:
- dropped the auth.rs and create.rs changes
- kept the scope to the API test
- renamed source/target to project_a/project_b
- replaced TestClient with AsyncOpenStack::authorize

gtema (changes requested):
let's now fix the PR subject and the summary to match what the change is actually doing

gtema (approved):

PR #908, opened 2026-07-03
feat(api): capture client IP via proxy headers & SPIFFE (#358 follow-up)

magdang:
Follow-up to #842 for the reopened #358. #842 captured the raw TCP peer on the public HTTP listener; per the maintainer comment two gaps remained (proxy header parsing, SPIFFE), while UDS is acceptable to skip.

## Proxy header parsing (config-gated, off by default)

New `[oslo_middleware] enable_proxy_headers_parsing` flag mirroring upstream Python Keystone's oslo.middleware (`HTTPProxyToWSGI`), **off by default**. When enabled, a middleware on the **public** interface parses RFC 7239 `Forwarded` (preferred) then `X-Forwarded-For`, takes the leftmost originating client, and overwrites `ConnectInfo<SocketAddr>`. Every downstream consumer (the `request` tracing span's `client.addr`, the API-Key IP allowlist, any future IP-based login control) transparently sees the real client.

Enabling it asserts the immediate peer is a trusted proxy; left off by default, a deployment not behind such a proxy cannot be tricked into trusting a spoofed header. The layer is wired **only** on the public interface — never on the internal SPIFFE / admin listeners.

## SPIFFE internal interface (`client_info` was not covered)

The internal mTLS listener uses `hyper::service::service_fn`, which bypasses axum's make-service, so it never populated `ConnectInfo`. The accepted TCP peer address is now injected by hand alongside the existing SVID / `Interface` extensions, so `client.addr` is captured on the internal interface too. Extracted into a small `attach_request_context` helper for testability.

The admin **UDS** interface is intentionally left uncovered — a Unix socket has no meaningful `SocketAddr` (acceptable per the issue).

## Tests
- Proxy-header parsing units (IPv4, `ip:port`, bracketed/bare IPv6, RFC 7239 params, obfuscated identifiers, precedence) + in-process middleware tests driving the real `into_make_service_with_connect_info` path.
- A composed public-listener test proving proxy rewrite + trailing-slash normalization (#734) + connect-info all compose, and that the layer runs before routing/normalization.
- A `[oslo_middleware]` config deserialization test.
- A SPIFFE `attach_request_context` unit test asserting `ConnectInfo`/`Interface`/SVID are attached.

`cargo fmt --check` and `cargo clippy --lib --tests` clean; full keystone lib + config suites pass.


crates/config/src/lib.rs
```diff
@@ -186,6 +188,10 @@ pub struct Config {
     #[serde(default)]
     pub mapping: MappingProvider,
 
+    /// `[oslo_middleware]` configuration (proxy header parsing, issue #358).
```
gtema: you do not need to refer to the issue here. It make sense only for some critical behavior (vulnerabilities, bugs, etc), but not a regular feature trackers

gtema (changes requested):
While it is good that you add certain compliance to oslo_middleware and rely on it's configuration, it does not by default implement sufficient security:

Proper model: don't blindly trust leftmost entry from any peer just cuz flag on. Standard practice (nginx, oslo.middleware, Go's httputil, etc):

1. Trusted-proxy allowlist, not just boolean flag.
Config should hold list of trusted proxy CIDRs (trusted_proxies = [...]). Middleware only rewrites ConnectInfo if immediate TCP peer is in that list. Else ignore header entirely — untrusted client hitting listener directly could forge X-Forwarded-For and set whatever IP it wants (bypass IP allowlist, spoof audit logs).

2. Walk chain right-to-left, strip trusted hops.
X-Forwarded-For: client, proxy1, proxy2 — proxy2 (closest, direct peer) appended last. Correct algo:
- start from rightmost entry
- if it's a trusted proxy, pop it, check next-left
- first entry NOT in trusted set = real client
- leftmost entry alone (current impl) is trivially spoofable — attacker just prepends fake IP as first hop, no proxy needed to strip it since nothing validates chain length/authenticity of entries left of the real edge proxy.

3. Cap hop count / header length. Guard against unbounded comma list (DoS on parsing) — set max hops parsed (e.g. 5-10).

4. Same treatment for Forwarded header — parse for= chain right-to-left same way, by= optionally cross-checked against known proxy identity.

---

I would at least extend the oslo_middleware config with the 'trusted_proxies' to do more secure extraction

crates/config/src/lib.rs
magdang: Done — removed the issue reference from the doc comment (and the other gratuitous `#358` references in the touched comments).

magdang:
Thanks — you were right that the boolean flag plus leftmost-`X-Forwarded-For` was spoofable. Reworked the extraction to the trusted-proxy model:

1. **Trusted-proxy allowlist** — added `[oslo_middleware] trusted_proxies` (CIDR list). The forwarding header is honoured only when the immediate TCP peer is within it; an empty list (default) trusts no one, so the raw peer is always kept.
2. **Right-to-left walk** — the effective client is now the rightmost address in the chain that is not itself a trusted proxy, so a prepended spoofed leftmost entry is never taken.
3. **Hop cap** — parsing is bounded to the rightmost 10 hops to guard against an unbounded comma list.
4. **`Forwarded` too** — the same right-to-left `for=` walk is applied to the RFC 7239 header.

As you noted this correlates with the API-key ingress, so rather than a second near-identical resolver I extracted the logic into a shared `core::api::forwarded::resolve_client_ip` that both the SCIM ingress and this middleware call; the shared version also picked up `Forwarded` support and the hop cap. `ConnectInfo` is only overwritten when a *different* upstream client is recovered, so a direct client keeps its real source port. Tests cover the peer-trust gate, the right-to-left walk, spoofed-leftmost, `Forwarded`, IPv6/port, and the hop cap.

crates/core/src/api/forwarded.rs
```diff
@@ -0,0 +1,261 @@
...
+/// are ignored entirely and the peer is returned unchanged.
+///
+/// `trusted_proxies` is a list of CIDR blocks (e.g. `10.0.0.0/8`); unparsable
+/// entries are skipped. Returns `None` only when no peer address is available.
+pub fn resolve_client_ip(
+    headers: &HeaderMap,
+    peer_ip: Option<IpAddr>,
+    trusted_proxies: &[String],
+) -> Option<IpAddr> {
+    let trusted: Vec<IpNet> = trusted_proxies
```
gtema: Now it becase visible and obvious, that trusted proxy IPs parsing is done on every request - this is inefficient. Instead it should be done during config reload. This can be actually also done by implementing a parser similar to the csv which is not only splitting string, but also immediately parses the string into IpAddr

gtema (changes requested):
Almost there

crates/core/src/api/forwarded.rs
magdang: Fixed. The trusted-proxy CIDRs are now parsed once at configuration-load time: a `csv_ipnet` deserializer (alongside the existing `csv`) splits the comma-separated value and parses each entry straight into `IpNet`, exactly as you suggested, so the request path only borrows a pre-parsed `&[IpNet]` and a malformed CIDR fails config loading up front.

While rebasing onto `main` this surfaced an overlap with `crate::net` (the XFF-only resolver rate-limiting and dynamic-plugin dispatch grew in the meantime), so I went one step further and unified them: `api::forwarded` is folded into `crate::net`, and all four ingress paths (API-Key, rate limiter, plugin dispatch, oslo_middleware layer) now share one resolver honoring RFC 7239 `Forwarded` → XFF fallback — which also brings the rate limiter and plugin dispatch in line with ADR 0025's "`X-Forwarded-For`/`Forwarded`" wording, and applies the parse-at-load fix to *every* `trusted_proxies` list, not just the new ones. ADR 0021/0022 wording updated accordingly.

gtema:
@ymh1874 you have a merge conflict here now

gtema (changes requested):
Two CONFIRMED bugs, both security-relevant:

1. spiffe_tls.rs:149 — internal mTLS listener now populates ConnectInfo(peer_addr) (verified absent pre-branch via git show origin/main:...). Three downstream checks assumed no-peer==internal-bypass and weren't updated: rate limiter, API-key IP allowlist, auth-plugin per-source limiter. Internal mesh traffic can now trip/bypass checks meant for public ingress only.
2. net.rs:105 — new Forwarded header trust has header-level precedence over XFF, with no check that the trusted proxy actually emits/strips Forwarded. Proxies trusted only for XFF (common default) may pass a client-forged Forwarded header straight through, letting it override the real XFF chain.

Plus 5 PLAUSIBLE/lower-severity: hop-cap window (>10 trusted hops), RFC 7239 quoted-comma mis-split, config load now fail-fast on unused malformed CIDR, implicit middleware-ordering dependency, and duplicated header-extraction logic (cleanup).

Bug 1:
Root cause — crates/keystone/src/server/listener/spiffe_tls.rs:149, attach_request_context:
```
extensions.insert(ConnectInfo(peer_addr));
```

This function runs for every request on the internal mTLS listener (Interface::Internal — the mesh-internal service-to-service listener, separate from the public HTTP listener and the UDS admin socket). Before this branch it only inserted the SPIFFE identity and Interface, never ConnectInfo. Verified with git show origin/main:crates/keystone/src/server/listener/spiffe_tls.rs — no ConnectInfo insert existed there. So this is a new behavior, added (per the comment) so client.addr shows up in audit logs for internal traffic too.

Why that's a problem — three consumers read ConnectInfo and treat its absence as "this is internal/admin traffic, don't apply public-ingress controls":

1. crates/core/src/rate_limit.rs — doc comment: "An absent peer identifies an internal/admin interface and bypasses this public-ingress limiter." Called unconditionally from crates/keystone/src/api/v3/auth/token/create.rs:118 via a PeerAddr extractor whose own doc comment still says "None on interfaces that don't populate ConnectInfo (e.g. the SPIFFE admin interface)" — now false for Interface::Internal.
2. crates/core/src/api/api_key_auth.rs:260-262 — effective_ip feeds ip_allowed(). With client_ip = None (old behavior) and an allowlist configured, ip_allowed always returned false (deny). Now client_ip = Some(mesh_peer_ip), so the check can pass or fail depending on whether that shared mesh IP happens to sit inside the configured allowlist.
3. crates/core/src/auth_plugin_auth.rs:132-144 (shown above) — check_per_source(remote_addr) bypasses (None => Ok(())) when there's no resolved address. Now remote_addr resolves to the shared internal mesh peer IP for every internal caller, so unrelated internal services hitting the same auth-plugin dispatch collapse into one check_per_source bucket.

Failure scenario — every internal-to-internal call (mesh sidecar terminates mTLS, so all internal callers likely share one peer IP) now: gets rate-limited by the public-ingress IP limiter as if it were external traffic; gets its API-key allowlist decision keyed on the shared mesh peer address instead of always-deny; and can trip/exhaust the per-source auth-plugin budget for every other internal caller sharing that peer IP — a noisy-neighbor DoS between internal services that couldn't happen pre-branch.

None of the three consumers were changed in this diff to gate on Interface::Internal explicitly — they still rely on peer-presence alone, which this branch quietly broke.

magdang:
Thanks for the detailed review. I fixed both confirmed bugs and reviewed all five lower severity findings.
 
Internal and admin requests now use an interface-aware peer check. ConnectInfo remains available for SPIFFE audit logs, but public rate limiting is skipped, API key IP rules remain fail closed, and auth
plugin calls no longer share one mesh peer bucket.
  
Proxy header trust is now explicit for each trust boundary. X-Forwarded-For is the default, and Forwarded requires explicit setup. Only the selected header is read. The parser handles quoted values and
rejects malformed, oversized, and over-limit chains as a whole.

The raw TCP peer is preserved before middleware rewriting, duplicate header parsing was removed, and invalid CIDRs still fail during config loading by design.

gtema (approved):

PR #910, opened 2026-07-03
Moves health and metrics endpoints to dedicated listener on separate port

### LINKED ISSUE #719: Move health endpoint to a dedicated app
Having health endpoint on a public app is potentially a ddos door. It should be instead moved to a dedicated app where metrics could be also added in future

mcblake27:
Health and metrics endpoints were previously attached to the public API router, allowing end users to access /health, /ready, and /metrics directly, which is a potential DDoS vector. This change moves them to a dedicated listener on port 8082, separate from the public API on port 8080. Also updates relevant tests to work with new structure/naming schema. Adds modular spawn_metrics_listener function to crates/keystone/src/bin/keystone.rs and API interface structure MetricsInterface in crates/config/src/interface.rs, as well as updating pathways in crates/config/src/lib.rs, crates/keystone/src/api/mod.rs, and crates/keystone/src/api/health.rs.

Closes #719.

This contribution was developed with the assistance of AI tools.

### DIFF
--- a/crates/config/src/interface.rs
+++ b/crates/config/src/interface.rs
@@ -66,6 +66,26 @@ impl Default for PublicInterface {
     }
 }
 
+/// Keystone health/metrics API interface.
+#[derive(Debug, Deserialize, Clone)]
+pub struct MetricsInterface {
+    /// Default address for the health/metrics endpoint. Defaults to `0.0.0.0:8082`.
+    #[serde(default = "default_metrics_tcp_address")]
+    pub tcp_address: SocketAddr,
+}
+
+impl Default for MetricsInterface {
+    fn default() -> Self {
+        Self {
+            tcp_address: default_metrics_tcp_address(),
+        }
+    }
+}
+
+fn default_metrics_tcp_address() -> SocketAddr {
+    SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), 8099)
+}
+
 fn default_public_tcp_address() -> SocketAddr {
     SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), 8080)
 }
--- a/crates/config/src/interface.rs
+++ b/crates/config/src/interface.rs
@@ -69,7 +69,7 @@ impl Default for PublicInterface {
 /// Keystone health/metrics API interface.
 #[derive(Debug, Deserialize, Clone)]
 pub struct MetricsInterface {
-    /// Default address for the health/metrics endpoint. Defaults to `0.0.0.0:8082`.
+    /// Default address for the health/metrics endpoint. Defaults to `0.0.0.0:8099`.
     #[serde(default = "default_metrics_tcp_address")]
     pub tcp_address: SocketAddr,
 }
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -257,7 +257,7 @@ pub struct Config {
     /// Reverse proxies trusted by the global per-IP rate limiter.
     #[serde(rename = "rate_limit_trusted_proxies", default)]
     pub rate_limit_trusted_proxies: RateLimitTrustedProxiesSection,
-    
+
     /// Server listener configuration for the health/metrics interface.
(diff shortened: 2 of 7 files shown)


crates/config/src/lib.rs
```diff
@@ -190,14 +190,18 @@ pub struct Config {
     #[serde(rename = "interface_internal", default)]
     pub interface_internal: Option<InternalInterface>,
 
-    /// Server listener configuration for the internal interface.
+    /// Server listener configuration for the public interface.
```
gtema, line 240: hehe, nice spot

crates/keystone/src/bin/keystone.rs
gtema, line 1236: this is the metrics handler that you need to merge with the health endpoint so that at the new listener we have 
- /health - the health endpoint
- /metrics - prometheus metrics

In one of the next steps we would evaluate how to populate metrics endpoint with the real data.

crates/config/src/interface.rs
```diff
@@ -66,6 +66,26 @@ impl Default for PublicInterface {
...
+impl Default for MetricsInterface {
+    fn default() -> Self {
+        Self {
+            tcp_address: default_metrics_tcp_address(),
+        }
+    }
+}
+
+fn default_metrics_tcp_address() -> SocketAddr {
+    SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), 8082)
```
gtema: now that you moved the health endpoint to the different address the k3s test fails since the keystone statefulset in kubernetes cannot reach it: https://github.com/openstack-experimental/keystone/blob/main/tools/k8s/keystone/base/statefulset-rs.yaml#L47C1-L72C1 - that place need to be adapted to point at the new place. Keep in mind the potential port conflict since keystone operates with many other components and systems (it does not conflict now, but any decision on changing ports is tricky - keycloak is also running on 8082, just in the different container). Let's maybe use 8099 as something less standard

gtema (changes requested):
I have updated the original issue with bit more details of the target setup how the new endpoint should work and combine healthz with otel metrics. Part of this change you do not need to implement OTEL service, but build an endpoint that will be further extended with it. Otherwise the change goes in the correct direction.
You need to fix commit message, formatting, and once you fix the deployment manifest the k3s test should also pass

crates/keystone/src/bin/keystone.rs
mcblake27: The way I am currently handling this is that when I define the router metrics_app in spawn_metrics_listener, I call .route("/metrics", axum::routing::get(metrics_handler)) to add this metrics handler as a route on the router. Is that satisfactory for what you're describing here, or is there a different way I am supposed to be doing this? My understanding is that that is what you are asking, but I may be misunderstanding.

crates/config/src/interface.rs
```diff
@@ -66,6 +66,26 @@ impl Default for PublicInterface {
     }
 }
 
+/// Keystone health/metrics API interface.
+#[derive(Debug, Deserialize, Clone)]
+pub struct MetricsInterface {
+    /// Default address for the health/metrics endpoint. Defaults to `0.0.0.0:8082`.
```
gtema: the docstring still says 8082 while the code does 8099

tools/k8s/keystone/base/statefulset-rs.yaml
```diff
@@ -43,19 +43,22 @@ spec:
             - containerPort: 8300
               name: keystone-raft
               protocol: TCP
+            - containerPort: 8099
+              name: keystone-metrics
```
gtema: as the k3s job log says kubernetes has length limitations (15 chars). Let's just use "metrics" here

gtema (changes requested):
few nits, otherwise good

gtema:
@mcblake27 please sync your fork repository so that it can be rebased solving the merge conflict

mcblake27:
@gtema Apologies, was having rebasing issues on my end but I believe it is okay now, let me know if there are still issues

gtema:
@mcblake27 format failure is genuine

gtema (approved):

Issue #369, opened 2025-11-16
Wrap credentials with the secrecy crate


gtema:
To prevent accidental logging/tracing of the secrets (token, password, application credential secret) we should wrap them with the secrecy crate like in the rust sdk

gtema:
currently at least the password is passed through the chain as a plain string

PR #912, opened 2026-07-03
feat(security): Wrap secrets with secrecy crate (#369)

magdang:
## TL;DR

Wraps every remaining plaintext secret in the codebase — the OIDC
`client_secret`, user/auth **passwords** (all the way up to the wire), the config
`invalid_password_hash_key`, and the OIDC token-exchange **bearer tokens** — in
`secrecy::SecretString`, so they can never leak through `Debug`,
`#[tracing::instrument]`, or accidental serialization. Secrets that must appear in
OPA policy / audit payloads are redacted to `"[REDACTED]"` instead of exposed.
Along the way this **fixes three pre-existing plaintext leaks into OPA policy
input** and closes an **empty-password acceptance gap**. The public OpenAPI schema
is byte-for-byte unchanged, and the change is covered by an adversarial
redaction/breakage test suite. Part of #369; delivered as one PR because touching
`api-types` pulls through every downstream crate.

---

## Background

`secrecy` was already a workspace dependency with an established pattern
(`database.connection: SecretString`), and passwords were already wrapped at the
provider/backend layers. This PR closes the **remaining gaps** the audit found and
lifts the wrapping to a consistent edge-to-consumer boundary: a secret is
`SecretString` from the moment it is deserialized off the wire (or read from the
DB) until the single point where it is genuinely consumed (hash / verify / HTTP
form / DB write), where it is explicitly `.expose_secret()`'d.

## What is wrapped (edge → consumer)

| Secret | Layers touched |
|---|---|
| **OIDC `oidc_client_secret`** | `api-types` DTO → `core-types` (3 structs) → SQL driver conversions → token-exchange consumer. `oidc_client_id` stays plaintext (per maintainer). |
| **User / auth passwords** | `UserCreate` / `UserUpdate` / `UserPassword` + the full nested `AuthRequest` tree, and the mirror `core-types` domain structs, are now `SecretString` from deserialize down. Exposure happens only at `hash_password` / `verify_password` / `validate_password`. |
| **Config `invalid_password_hash_key`** | `SecurityComplianceProvider` (mirrors `database.connection`). |
| **OIDC `id_token` / `access_token`** | `TokenExchangeResponse` bearer tokens, previously plaintext `String` in a `Debug`-deriving struct; exposed only at the JWT-verify boundary. |

### Redaction, not exposure
`SecretString` intentionally does not implement `Serialize`. For the DTOs that
*must* stay serializable (they are embedded in OPA policy input / audit payloads),
a `serialize_with` helper renders a present secret as a fixed `"[REDACTED]"`
marker — field presence is preserved, the value never leaves. This is deliberately
different from the existing "expose once" helper used for the one-time API-key
token (that one is part of the API contract).

## Pre-existing leaks this fixes

`json!({"identity_provider": current})`, `json!({"user": req.user})` (and the
create/update variants) serialize the domain/DTO structs straight into **OPA
policy input**. Before this PR those payloads contained the **plaintext**
`oidc_client_secret` and user `password`. They now redact.

## Security hardening — empty-password guard

Wrapping the password meant the per-DTO `length(min = 1)` validation could not be
kept: validator 0.20's `custom` unconditionally calls
`ValidationError::add_param(&field)`, which requires the field to be `Serialize` —
and secrets deliberately are not. Without that guard, on a default deployment
(no `password_regex`) an **empty password would be accepted and stored** (`bcrypt`
hashes an empty string fine).

Fixed by moving the non-empty check into the correct central layer —
`SecurityComplianceProvider::validate_password` — which runs on the wrapped value
for **every** write path (create / update / change). This is strictly more secure
than before (previously only create was guarded).

## Design decisions / notes

- **Sea-ORM `oidc_client_secret` column stays plaintext `String`.** Wrapping the
  DB column would require custom `TryGetable` / `Into<Value>` impls and would not
  stop SQL-parameter-level logging anyway; the DB write is the final consuming
  method, where the secret is unwrapped.
- **`PartialEq` / `Default`** are dropped from the wrapped structs where nothing
  relies on them (matching the existing app-cred / k8s pattern), and
  re-implemented **manually** where tests do rely on them (`IdentityProvider`
  read model; `UserPasswordAuthRequest`), preserving the prior public contract.
- **OpenAPI unchanged.** `#[schema(value_type = String / Option<String>)]` keeps
  every wrapped field documented exactly as before; the generated spec has **zero
  structural diff** vs `main`.
- **validator + `SecretString` are fundamentally incompatible** for field-level
  length/custom validation (see the empty-password section) — documented inline so
  a future pass doesn't reintroduce it.

## Testing & verification

- `cargo check --workspace --all-targets --all-features` ✅
- `cargo clippy --lib --tests --all-features` ✅ · `cargo fmt --check` ✅
- Full unit/lib suite green (no regressions): api-types, core-types, config, core,
  keystone, and the SQL drivers.
- **OpenAPI**: regenerated and diffed against `main` — **0 structural changes**.
- **Adversarial redaction / breakage test suite** added, covering:
  - `#[serde(flatten)] extra` + custom `serialize_with` + `skip` on `UserCreate` /
    `UserUpdate` (the exact struct serialized to OPA) — password redacted, `extra`
    preserved, via both `to_string` and `to_value`.
  - Deep-nested `AuthRequest` serialize redaction (password 4 levels deep).
  - OPA-path redaction for the user + IdP create/update DTOs.
  - Empty-password rejected without a regex (regression guard).
  - `Debug` redaction matrix across every wrapped type, incl. the config hash key.
  - IdP response DTO can never expose a client secret.
  - OIDC bearer-token `Debug` redaction; deserialize edge cases (null / absent).
- Every `.expose_secret()` call site re-audited — none feeds a log / print /
  format; each is a genuine consume boundary (hash, verify, DB write, HTTP form,
  JWT verify).

## Out of scope (follow-up #370)

The credential `blob` (EC2 secret key / TOTP seed material) is also a secret but is
outside the maintainer's stated scope for this effort; left as a follow-up so it
isn't silently dropped.


crates/api-types/src/federation/identity_provider.rs
```diff
@@ -169,12 +174,21 @@ pub struct IdentityProviderCreate {
...
 
     /// The oidc `client_secret` to use for the private client. It is never
-    /// returned back.
-    #[cfg_attr(feature = "builder", builder(default))]
-    #[cfg_attr(feature = "openapi", schema(nullable = false))]
-    #[serde(skip_serializing_if = "Option::is_none")]
-    #[cfg_attr(feature = "validate", validate(length(max = 255)))]
-    pub oidc_client_secret: Option<String>,
+    /// returned back. Wrapped in [`SecretString`] to prevent accidental
+    /// exposure via Debug/tracing; redacted (never exposed) when serialized into
```
gtema: I think you should not add comment "Wrapped in [`SecretString`] to prevent accidental
    /// exposure via Debug/tracing; redacted (never exposed) ...." everywhere. The purpose is self explaining. I guess all new docstrings changes are not really necessary

crates/api-types/src/federation/identity_provider.rs
```diff
@@ -383,3 +411,56 @@ pub struct IdentityProviderListParameters {
...
+    #[test]
+    fn idp_create_redacts_client_secret_for_policy() {
+        let create: IdentityProviderCreate =
+            serde_json::from_str(r#"{"name":"idp","oidc_client_secret":"CSLEAK"}"#).unwrap();
+        let rendered = serde_json::json!({ "identity_provider": &create }).to_string();
+        assert!(
+            !rendered.contains("CSLEAK"),
+            "leaked client secret: {rendered}"
+        );
+        assert!(rendered.contains("[REDACTED]"), "not redacted: {rendered}");
```
gtema: this opens doors to way too many fixes in unittests should the 'secrecy' crate change the content. Just a check for not containing plain text secret should be sufficient

crates/api-types/src/v3/auth/token.rs
gtema: this should be hidden as well

gtema: this is also a secret

crates/api-types/src/secret_serde.rs
```diff
@@ -0,0 +1,157 @@
...
+// NOTE: length/non-empty validation for wrapped secrets is deliberately NOT done
+// with validator's field-level `custom` here: validator 0.20 unconditionally
+// calls `ValidationError::add_param(&field)`, which requires the field to be
+// `Serialize` — and `SecretString` intentionally is not. Password non-emptiness
+// is instead enforced centrally in
+// `openstack_keystone_config::SecurityComplianceProvider::validate_password`,
+// which runs on the wrapped value at the service layer for every write path.
+
+/// Redact an `Option<SecretString>` on serialize.
+pub(crate) fn serialize_secret_redacted<S>(
```
gtema: ehm, I am actually not sure you need this. Have a look at K8sAuthRequest. Maybe I am wrong though

crates/core-types/src/federation/identity_provider.rs
```diff
@@ -14,13 +14,32 @@
...
-use serde::Serialize;
+use secrecy::{ExposeSecret, SecretString};
+use serde::{Serialize, Serializer};
 use serde_json::Value;
 
 use crate::error::BuilderError;
 
+/// Serialize an optional secret as a fixed redaction marker so that it never
+/// leaks in Debug/policy/audit payloads while still signalling presence.
+fn serialize_secret_redacted<S>(
```
gtema: now again the nearly same function as above

crates/core-types/src/federation/identity_provider.rs
```diff
@@ -81,8 +104,42 @@ pub struct IdentityProvider {
     pub provider_config: Option<Value>,
 }
 
+/// Manual `PartialEq` (the derive cannot be used because `oidc_client_secret`
+/// is a [`SecretString`], which does not implement `PartialEq`). Preserves the
+/// pre-wrapping equality contract by comparing the exposed secret values.
+impl PartialEq for IdentityProvider {
```
gtema: This is bad. I fear when we add new field to any struct that manually overrides PartialEq we definitely forget to extend the check here. Please investigate for another solution - it is just a door for bugs

gtema (changes requested):

crates/api-types/src/federation/identity_provider.rs
magdang: Agreed — dropped the "Wrapped in `SecretString` …" docstrings throughout. Redaction is now a property of the `RedactedSecret` type itself, so the fields just carry their functional description.

crates/api-types/src/federation/identity_provider.rs
magdang: Done — these redaction tests now assert only that the plaintext secret is absent from the rendered output; I removed the `contains("[REDACTED]")` assertions here and in the sibling create/update tests so nothing depends on the exact marker text.

crates/api-types/src/v3/auth/token.rs
magdang: Done — `passcode` is now a `RedactedSecret` in both the api-types DTO and the core-types `UserTotpAuthRequest`, so it no longer leaks via Debug/serialization. `verify_totp` reads it through `expose_secret()` at the point of use.

magdang: Done — `TokenAuth.id` is now a `RedactedSecret`; it is exposed only at the `authorize_by_token` call boundary via `expose_secret()`.

crates/api-types/src/secret_serde.rs
magdang: You were right. I removed `secret_serde.rs` entirely. Rather than a module of serialize-helpers, redaction is now built into the `RedactedSecret` newtype, so DTOs just `#[derive(Serialize)]` with no per-field glue. (`K8sAuthRequest` intentionally stays on `SecretString` + its local serializer — it serializes the JWT *exposed* to forward it, the opposite requirement, so it should not use the redacting type.)

crates/core-types/src/federation/identity_provider.rs
magdang: Removed — there is no longer a per-crate serializer to duplicate. `RedactedSecret` is defined once in `core-types` and api-types now depends on it directly, so both layers share the same type and its single Serialize impl.

crates/core-types/src/federation/identity_provider.rs
magdang: Agreed, that hand-written impl was fragile for exactly that reason. `RedactedSecret` implements `PartialEq` (comparing the underlying values), so `IdentityProvider` derives `PartialEq` normally again — a newly added field is now automatically included in the comparison and the manual impl is gone. Same approach lets `UserTotpAuthRequest` keep its derived `PartialEq`/`Default`.

magdang:
Thanks for the review. I reworked the secret handling around a single `RedactedSecret` newtype in `core-types` (redacts on Debug/Serialize, derives `PartialEq`/`Serialize`/`Deserialize` on the owning struct). That let me:

- delete `secret_serde.rs` and the duplicated core-types serializer,
- drop the hand-written `PartialEq for IdentityProvider` (it derives again, so new fields are picked up automatically),
- wrap the remaining plaintext secrets (`passcode`, `TokenAuth.id`),
- trim the repetitive docstrings and relax the redaction tests to only assert plaintext absence.

`api-types` now depends on `core-types` directly (previously optional) so both layers share the type; every existing consumer already enabled `builder`/`conv`, so the only new edge is `cli-manage` picking up the pure-types crate. Individual replies inline. `cargo build`/`clippy --lib --tests`/tests all green.

crates/api-types/Cargo.toml
```diff
@@ -22,7 +22,7 @@ chrono = { workspace = true, default-features = false, features = ["serde"] }
 derive_builder = { workspace = true, optional = true }
 eyre.workspace = true
 http = { workspace = true, optional = true }
-openstack-keystone-core-types = { workspace = true, optional = true }
```
gtema, line 25: This is for sure undesired, this pulls lots of very heavy dependencies to every crate that is only interested in API types

gtema (changes requested):

crates/core-types/src/secret.rs
```diff
@@ -0,0 +1,129 @@
...
+/// over the wire (e.g. a password in `POST /v3/auth/tokens`). This mirrors the
+/// `K8sAuthRequest` handling. It does **not** redact on serialize; secrecy's
+/// contract is `Debug`-redaction, which is what protects logs and traces. The
+/// only server-side serialization of these fields is into the local, trusted
+/// OPA policy subprocess, which is inside the trust boundary; client-facing
+/// response DTOs never carry a secret field.
+#[derive(Clone, Default)]
+pub struct SecretField(SecretString);
+
+impl fmt::Debug for SecretField {
```
gtema: I also do not really get why you do this. SecretString already prevents exposure of the content which is why the crate is used

crates/api-types/Cargo.toml
magdang: Agreed, that was wrong. Reverted it: core-types is optional again and gated under builder/conv like before, so a plain api-types consumer does not pull it in. The DTOs use secrecy::SecretString directly now, which api-types already had as a dependency, so nothing heavy gets added.

crates/core-types/src/secret.rs
magdang: You are right, dropped the newtype. The fields are just SecretString now with a small serialize_with helper for transport, same as K8sAuthRequest. The only thing the wrapper bought me was a derivable PartialEq on IdentityProvider, but K8sAuthRequest does not derive PartialEq either, so I removed it from the secret-bearing structs and switched the two driver tests that compared the whole struct to compare fields instead. No manual PartialEq anywhere.

crates/core/src/identity/service.rs
```diff
@@ -438,7 +438,13 @@ impl IdentityApi for IdentityService {
                 .and_then(serde_json::Value::as_u64)
                 .map(|d| d as u32)
                 .unwrap_or(30);
-            crate::credential::totp::verify_totp(seed, &auth.passcode, digits, period, now)
+            crate::credential::totp::verify_totp(
+                seed,
+                auth.passcode.expose_secret(),
```
gtema: it would be good if we can expose the secret as late as possible (within the verify_totp)

crates/api-types/src/federation/identity_provider.rs
```diff
@@ -12,13 +12,46 @@
...
 #[cfg(feature = "validate")]
 use validator::Validate;
 
 use crate::Link;
 
+/// Serialize an optional client secret transparently for transport (the client
+/// must send the real value). `Debug` redaction from `SecretString` is what
+/// keeps it out of logs; the value is never serialized back to a client because
+/// the response type has no secret field.
+fn serialize_optional_secret<S>(
```
gtema: can you please move all "serialize_xxx_secret" helpers used across the crate to the common.rs. Right now there is a code duplication for those.

crates/api-types/src/v3/auth/token.rs
```diff
@@ -335,8 +348,9 @@ pub struct User {
 #[cfg_attr(feature = "validate", derive(validator::Validate))]
 pub struct TokenAuth {
     /// An authentication token.
-    #[cfg_attr(feature = "validate", validate(length(max = 1024)))]
-    pub id: String,
+    #[cfg_attr(feature = "openapi", schema(value_type = String))]
```
gtema, line 371: We need to find a way to keep the validation for the wrapped secrets

crates/api-types/src/v3/user.rs
gtema, line 220: Can you please implement "impl UserCreate {  fn to_policy_input(&self) -> serde_json::Value {:....." for UserCreate, UserUpdate, IdentityProviderCreate, IdentityProviderUpdate and in the http handlers use it when the object is being passed to the policy engine: ` json!({"user": req.user.to_policy_input()})` so that we do not leak plain text secrets into the policy engine (which actually goes through the network).

gtema (changes requested):

crates/core/src/identity/service.rs
magdang: Done. `verify_totp` now takes the wrapped passcode and exposes it internally, so the plaintext is revealed as late as possible.

crates/api-types/src/federation/identity_provider.rs
magdang: Done. The shared secret serializers are now centralized in `api-types/src/common.rs` and reused by the API DTOs.

crates/api-types/src/v3/auth/token.rs
magdang: Done. Validation is preserved with schema validators for wrapped secrets, avoiding serialization of the secret into validator params.

crates/api-types/src/v3/user.rs
magdang: Done. Added `to_policy_input()` for user and identity provider create/update DTOs, and updated the handlers to pass redacted values to OPA.

crates/api-types/src/federation/identity_provider.rs
```diff
@@ -239,8 +247,73 @@ pub struct IdentityProviderCreate {
...
+        if let Some(value) = &self.domain_id {
+            input.insert("domain_id".to_string(), serde_json::json!(value));
+        }
+        if let Some(value) = &self.oidc_discovery_url {
+            input.insert("oidc_discovery_url".to_string(), serde_json::json!(value));
+        }
+        if let Some(value) = &self.oidc_client_id {
+            input.insert("oidc_client_id".to_string(), serde_json::json!(value));
+        }
+        if self.oidc_client_secret.is_some() {
```
gtema: this is not necessary and makes no sense (passing masked input where it should be skipped completely). Technically I would rather convert the whole struct to Value and remove secrets out of it. It is much shorter. The same would apply to every other case

crates/api-types/src/federation/identity_provider.rs
```diff
@@ -133,6 +134,10 @@ pub struct IdentityProviderResponse {
 )]
 #[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
 #[cfg_attr(feature = "validate", derive(validator::Validate))]
+#[cfg_attr(
+    feature = "validate",
```
gtema, line 138: problem is that now you ONLY validate a single field dropping validation of all other fields. Instead you can attach a specific validation function to a field directly.

crates/api-types/src/federation/identity_provider.rs
```diff
@@ -251,6 +324,10 @@ pub struct IdentityProviderCreate {
...
+    validate(schema(function = "validate_identity_provider_update_secret"))
```
gtema, line 286: same issue with the validation as above

crates/api-types/src/federation/identity_provider.rs
```diff
@@ -383,3 +528,84 @@ pub struct IdentityProviderListParameters {
...
+
+        let update: IdentityProviderUpdate =
+            serde_json::from_str(r#"{"oidc_client_secret":"CSLEAK2"}"#).unwrap();
+        let input = update.to_policy_input();
+        let rendered = input.to_string();
+        assert!(
+            !rendered.contains("CSLEAK2"),
+            "policy input leaked client secret: {rendered}"
+        );
+        assert_eq!(
```
gtema: as mentioned above - the field should simply not appear at all in the policy data

crates/api-types/src/v3/auth/token.rs
```diff
@@ -227,6 +228,10 @@ pub struct PasswordAuth {
 )]
 #[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
 #[cfg_attr(feature = "validate", derive(validator::Validate))]
+#[cfg_attr(
+    feature = "validate",
```
gtema, line 232: same issue with broken validation

crates/api-types/src/v3/auth/token.rs
```diff
@@ -275,6 +286,10 @@ pub struct TotpAuth {
...
```
gtema, line 294: and here

crates/api-types/src/v3/auth/token.rs
```diff
@@ -333,10 +354,20 @@ pub struct User {
 )]
 #[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
 #[cfg_attr(feature = "validate", derive(validator::Validate))]
+#[cfg_attr(
```
gtema, line 365: and here

crates/api-types/src/v3/auth_conv.rs
```diff
@@ -41,7 +42,7 @@ impl TryFrom<api_types::UserPassword>
             }
             upa.domain(domain_builder.build()?);
         }
-        upa.password(value.password.clone());
+        upa.password(value.password.expose_secret());
```
gtema: we should not expose secret here, but pass it down. This signals that upa.password must also be converted to SecretString

crates/api-types/src/v3/user.rs
```diff
@@ -113,6 +114,10 @@ pub struct UserResponse {
 )]
 #[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
 #[cfg_attr(feature = "validate", derive(validator::Validate))]
+#[cfg_attr(
+    feature = "validate",
```
gtema, line 118: validation

crates/api-types/src/v3/user.rs
```diff
@@ -182,6 +221,10 @@ pub struct UserCreateRequest {
...
```
gtema, line 217: and here

crates/core-types/src/identity/user.rs
```diff
@@ -15,12 +15,37 @@ use std::collections::HashMap;
...
 use derive_builder::Builder;
+use secrecy::{ExposeSecret, SecretString};
 use serde::Serialize;
 use serde_json::Value;
-use validator::Validate;
+use validator::{Validate, ValidationError};
 
 use crate::error::BuilderError;
 
+fn validate_secret_length(secret: &SecretString, max: usize) -> Result<(), ValidationError> {
```
gtema, line 25: also this validators can be placed to common for reuse

crates/core-types/src/identity/user.rs
```diff
@@ -75,7 +100,11 @@ pub struct UserResponse {
 }
 
 /// User creation data.
-#[derive(Builder, Clone, Debug, PartialEq, Validate)]
+///
+/// `PartialEq` is intentionally not derived: `password` is wrapped in
+/// [`SecretString`], which does not implement `PartialEq` by design.
+#[derive(Builder, Clone, Debug, Validate)]
+#[validate(schema(function = "validate_user_create_secret"))]
```
gtema, line 107: again validation

gtema (changes requested):

crates/api-types/src/federation/identity_provider.rs
magdang: Done in `7c360743` — `to_policy_input()` now omits the secret entirely instead of masking it with `[REDACTED]`, for `UserCreate/Update` and `IdentityProviderCreate/Update`.

crates/api-types/src/federation/identity_provider.rs
magdang: As above — the secret key no longer appears in the policy data.

crates/api-types/src/federation/identity_provider.rs
magdang: A field-level `#[validate(custom)]` can't be attached to a `SecretString`: validator 0.20 calls `err.add_param("value", &field)` on failure, which requires `Serialize` — it won't compile and would serialize the secret into the error. So I kept `#[derive(Validate)]` (every non-secret field is still validated) and check only the secret via a struct-level schema fn that never serializes it, documented with a `NOTE` inline. Added a test that an over-long non-secret field (e.g. `name`) still fails. `7c360743`

crates/api-types/src/federation/identity_provider.rs
magdang: As above.

crates/api-types/src/v3/auth/token.rs
magdang: As above.

crates/api-types/src/v3/auth/token.rs
magdang: As above.

crates/api-types/src/v3/auth/token.rs
magdang: As above.

crates/api-types/src/v3/user.rs
magdang: As above.

crates/api-types/src/v3/user.rs
magdang: As above.

crates/core-types/src/identity/user.rs
magdang: As above.

crates/api-types/src/v3/auth_conv.rs
magdang: Done in `7c360743` — the conversion passes the `SecretString` down instead of exposing it at conversion time.

crates/core-types/src/identity/user.rs
magdang: The `api-types` and `core-types` copies are separate because each crate gates validation behind its own `validate` feature, so a shared module would couple the feature sets. Happy to extract one if you'd prefer.

crates/api-types/src/v3/user.rs
magdang: Follow-up: the **v4** user create/update handlers were still passing the raw DTO to OPA (leaking the password); now routed through `to_policy_input()` too, with regression tests asserting no `password` key reaches policy input. `7c360743`

crates/api-types/src/v3/auth/token.rs
magdang: Final approach: kept struct-level schema validation for the secret (see the explanation on `identity_provider.rs`); field-level `custom` isn't possible on `SecretString`. `7c360743`

crates/api-types/src/federation/identity_provider.rs
gtema: hmm, ok. I didn't knew this specific of validator crate. It's comment state that the struct validator is called after all other individual fields are validated. Good.

crates/api-types/src/federation/identity_provider.rs
```diff
@@ -323,8 +402,70 @@ pub struct IdentityProviderUpdate {
     pub provider_config: Option<Option<Value>>,
 }
 
+impl IdentityProviderUpdate {
+    #[must_use]
+    pub fn to_policy_input(&self) -> serde_json::Value {
+        let mut input = serde_json::Map::new();
```
gtema: you decided not to dump to value and strip secret fields manually? This results it too much code and accidental miss when modifying params

crates/api-types/src/v3/user.rs
```diff
@@ -154,14 +159,49 @@ pub struct UserCreate {
...
+        default,
+        skip_serializing_if = "Option::is_none",
+        serialize_with = "crate::common::serialize_optional_secret"
+    )]
+    pub password: Option<SecretString>,
+}
+
+impl UserCreate {
+    #[must_use]
+    pub fn to_policy_input(&self) -> serde_json::Value {
```
gtema, line 176: also here it would be shorter and safer to dump the struct to Value and strip secrets

crates/core-types/src/identity/user.rs
gtema: those are just few lines, that's ok. At some point we may end up having "common" crate, but not now

crates/identity-driver-sql/src/authenticate.rs
```diff
@@ -79,9 +80,11 @@ pub async fn authenticate_by_password(
             .await
             .map_err(IdentityProviderError::password_hash)?;
 
-        let _ = password_hashing::verify_password(config, &auth.password, dummy_hash)
-            .await
-            .map_err(IdentityProviderError::password_hash)?;
+        // Exposure boundary: unwrapped only to feed the constant-time verify.
+        let _ =
+            password_hashing::verify_password(config, auth.password.expose_secret(), dummy_hash)
```
gtema: I actually also thought of passing the SecretString down to the final hasher, but I agree this is a bit debatable and cause too much mess. Let's leave it as a followup to define whether we do it or find a method to guarantee value is not logged by hasher

crates/keystone/src/api/v3/auth/token/common.rs
```diff
@@ -60,7 +61,7 @@ pub(super) async fn authenticate_request(
                 .get_token_provider()
                 .authorize_by_token(
                     &ExecutionContext::internal(state),
-                    &token.id,
+                    token.id.expose_secret(),
```
gtema: this is, however, very important - authorize_by_xxx methods have logging and also decorators eventually capturing parameters. We NEED to ensure we do not expose the secret on this level yet

crates/identity-driver-sql/src/authenticate.rs
gtema: later: since you would need to adapt the authenticate_by_token anyway - let's pass secret into the verify_password and hash_password and make them expose the secret when calling the hashers

crates/api-types/src/federation/identity_provider.rs
magdang: Done in `21b23e2a`. `to_policy_input` now serializes the whole struct with `serde_json::to_value(self)` and strips only the secret key (`oidc_client_secret`), so non-secret fields are never missed when params change. Applied to both the Create and Update variants.

crates/api-types/src/v3/user.rs
magdang: Done in `21b23e2a`. Same treatment here: dump the struct to `Value` and remove `password`. Shorter, and it can't silently drop a newly added field from policy input.

crates/identity-driver-sql/src/authenticate.rs
magdang: Done in `ffe3e603`. `hash_password` and `verify_password` now take `&SecretString` and call `expose_secret()` internally, right before dispatching to the concrete hasher. The identity and application-credential SQL drivers pass the wrapped secret down instead of exposing it at the call site.

crates/keystone/src/api/v3/auth/token/common.rs
magdang: Done in `fae12f4c`. `authorize_by_token` now takes the credential as `&SecretString`; it is exposed only inside the `TokenService` impl, immediately before `validate_to_context_impl`, so the plaintext never crosses the trait boundary where tracing/decorators run. Callers pass the wrapped secret (the v3 auth handler passes `&token.id`; the `X-Auth-Token` extractor wraps the header).

crates/keystone/src/federation/api/identity_provider/update.rs
```diff
@@ -70,8 +70,8 @@ pub(super) async fn update(
         .enforce(
             "identity/federation/identity_provider/update",
             &user_auth,
-            json!({"identity_provider": current}),
-            Some(json!({"identity_provider": req.identity_provider})),
+            json!({"identity_provider": req.identity_provider.to_policy_input()}),
```
gtema, line 73: good fix for ordering, but we also need to pass `current.to_policy_input` - let's do this in the followup. This had already too many iterations and is good this way

gtema (approved):

Issue #782, opened 2026-06-09
Expose Application Credential API


gtema:
https://docs.openstack.org/api-ref/identity/v3/index.html#application-credentials

 [ ] - api v3 (create/list/show/delete)
 [ ] - policy
 [ ] - api test

konac-hamza:
@gtema  The provider and backend level doesn't have any method for delete app-cred although pythonv3 code has. Should I skip delete api in the PR which I am working on for this issue or  should I implement missing methods in the PR?

gtema:
Implement delete.

PR #948, opened 2026-07-07
Feature/expose app credential

konac-hamza:
This PR implements the Application Credential API (part of #782).

### Changes

* Added API type definitions for `application_credential` and `access_rule`, along with their converters in the `api-types` crate.
* Followed the Python Keystone implementation to preserve compatibility and prevent known security issues in the endpoints.
* Implemented the delete operation in both the core and SQL layers.
* Implemented policy enforcement based on the default Keystone policies.

### DIFF
--- a/crates/api-types/src/error_conv.rs
+++ b/crates/api-types/src/error_conv.rs
@@ -279,7 +279,7 @@ impl From<ApplicationCredentialProviderError> for KeystoneApiError {
                 Self::BadRequest("application credential has expired".into())
             }
             ApplicationCredentialProviderError::AccessRuleInUse(x) => {
-                Self::BadRequest(format!("access rule {x} is still in use"))
+                Self::Conflict("application credential access rule is in use".into())
             }
             err @ ApplicationCredentialProviderError::AccessRulesUnenforced => {
                 Self::BadRequest(err.to_string())
--- a/crates/api-types/src/error_conv.rs
+++ b/crates/api-types/src/error_conv.rs
@@ -622,6 +622,40 @@ impl From<KeystoneError> for KeystoneApiError {
     }
 }
 
+impl From<ApplicationCredentialProviderError> for KeystoneApiError {
+    fn from(value: ApplicationCredentialProviderError) -> Self {
+        match value {
+            ApplicationCredentialProviderError::ApplicationCredentialNotFound(x) => {
+                Self::NotFound {
+                    resource: "application_credential".into(),
+                    identifier: x,
+                }
+            }
+            ApplicationCredentialProviderError::AccessRuleNotFound(x) => Self::NotFound {
+                resource: "access_rule".into(),
+                identifier: x,
+            },
+            ApplicationCredentialProviderError::RoleNotFound(x) => Self::NotFound {
+                resource: "role".into(),
+                identifier: x,
+            },
+            ref err @ ApplicationCredentialProviderError::Conflict(..) => {
+                Self::Conflict(err.to_string())
+            }
+            ref err @ ApplicationCredentialProviderError::Validation { .. } => {
+                Self::BadRequest(err.to_string())
+            }
+            ApplicationCredentialProviderError::ApplicationCredentialExpired => {
+                Self::BadRequest("application credential has expired".into())
+            }
+            ApplicationCredentialProviderError::AccessRuleInUse(x) => {
+                Self::BadRequest(format!("access rule {x} is still in use"))
+            }
+            other => Self::InternalError(other.to_string()),
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/crates/api-types/src/error_conv.rs
+++ b/crates/api-types/src/error_conv.rs
@@ -253,19 +253,33 @@ impl From<CatalogProviderError> for KeystoneApiError {
 }
 
 impl From<ApplicationCredentialProviderError> for KeystoneApiError {
-    fn from(source: ApplicationCredentialProviderError) -> Self {
-        match source {
+    fn from(value: ApplicationCredentialProviderError) -> Self {
+        match value {
             ApplicationCredentialProviderError::ApplicationCredentialNotFound(x) => {
                 Self::NotFound {
                     resource: "application_credential".into(),
                     identifier: x,
                 }
             }
+            ApplicationCredentialProviderError::AccessRuleNotFound(x) => Self::NotFound {
+                resource: "access_rule".into(),
+                identifier: x,
+            },
+            ApplicationCredentialProviderError::RoleNotFound(x) => Self::NotFound {
+                resource: "role".into(),
+                identifier: x,
+            },
             ref err @ ApplicationCredentialProviderError::Conflict(..) => {
                 Self::Conflict(err.to_string())
             }
-            err @ ApplicationCredentialProviderError::ApplicationCredentialExpired => {
-                Self::unauthorized(err, None::<String>)
+            ref err @ ApplicationCredentialProviderError::Validation { .. } => {
+                Self::BadRequest(err.to_string())
+            }
+            ApplicationCredentialProviderError::ApplicationCredentialExpired => {
+                Self::BadRequest("application credential has expired".into())
+            }
+            ApplicationCredentialProviderError::AccessRuleInUse(x) => {
+                Self::BadRequest(format!("access rule {x} is still in use"))
             }
             err @ ApplicationCredentialProviderError::AccessRulesUnenforced => {
                 Self::BadRequest(err.to_string())
@@ -618,40 +632,6 @@ impl From<KeystoneError> for KeystoneApiError {
             KeystoneError::TokenProvider { source } => source.into(),
             KeystoneError::TrustProvider { source } => source.into(),
             _ => KeystoneApiError::internal(value),
-        }
-    }
-}
-
-impl From<ApplicationCredentialProviderError> for KeystoneApiError {
-    fn from(value: ApplicationCredentialProviderError) -> Self {
-        match value {
-            ApplicationCredentialProviderError::ApplicationCredentialNotFound(x) => {
-                Self::NotFound {
-                    resource: "application_credential".into(),
-                    identifier: x,
-                }
-            }
-            ApplicationCredentialProviderError::AccessRuleNotFound(x) => Self::NotFound {
-                resource: "access_rule".into(),
-                identifier: x,
-            },
-            ApplicationCredentialProviderError::RoleNotFound(x) => Self::NotFound {
-                resource: "role".into(),
-                identifier: x,
-            },
-            ref err @ ApplicationCredentialProviderError::Conflict(..) => {
-                Self::Conflict(err.to_string())
-            }
-            ref err @ ApplicationCredentialProviderError::Validation { .. } => {
-                Self::BadRequest(err.to_string())
-            }
-            ApplicationCredentialProviderError::ApplicationCredentialExpired => {
-                Self::BadRequest("application credential has expired".into())
-            }
-            ApplicationCredentialProviderError::AccessRuleInUse(x) => {
-                Self::BadRequest(format!("access rule {x} is still in use"))
-            }
-            other => Self::InternalError(other.to_string()),
         }
     }
 }
--- a/crates/api-types/src/error_conv.rs
+++ b/crates/api-types/src/error_conv.rs
@@ -278,7 +278,7 @@ impl From<ApplicationCredentialProviderError> for KeystoneApiError {
             ApplicationCredentialProviderError::ApplicationCredentialExpired => {
                 Self::BadRequest("application credential has expired".into())
             }
-            ApplicationCredentialProviderError::AccessRuleInUse(x) => {
+            ApplicationCredentialProviderError::AccessRuleInUse(_) => {
                 Self::Conflict("application credential access rule is in use".into())
             }
             err @ ApplicationCredentialProviderError::AccessRulesUnenforced => {
--- a/crates/api-types/src/v3/application_credential/access_rule.rs
+++ b/crates/api-types/src/v3/application_credential/access_rule.rs
@@ -1,0 +1,88 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+use chrono::{DateTime, Utc};
+use serde::{Deserialize, Serialize};
+#[cfg(feature = "validate")]
+use validator::Validate;
+
+/// Short access rule representation.
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[cfg_attr(
+    feature = "builder",
+    derive(derive_builder::Builder),
+    builder(
+        build_fn(error = "crate::error::BuilderError"),
+        setter(strip_option, into)
+    )
+)]
+#[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
+#[cfg_attr(feature = "validate", derive(validator::Validate))]
+pub struct AccessRule {
+    /// The ID of the access rule.
+    #[cfg_attr(feature = "validate", validate(length(min = 1, max = 64)))]
+    pub id: String,
+
+    /// The HTTP method permitted.
+    #[cfg_attr(feature = "builder", builder(default))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[cfg_attr(feature = "validate", validate(length(min = 1, max = 16)))]
+    pub method: Option<String>,
+
+    /// The API path permitted.
+    #[cfg_attr(feature = "builder", builder(default))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[cfg_attr(feature = "validate", validate(length(min = 1, max = 128)))]
+    pub path: Option<String>,
+
+    /// The service type permitted.
+    #[cfg_attr(feature = "builder", builder(default))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[cfg_attr(feature = "validate", validate(length(min = 1, max = 64)))]
+    pub service: Option<String>,
+}
+
+/// Access rule for creation (id is optional).
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[cfg_attr(
+    feature = "builder",
+    derive(derive_builder::Builder),
+    builder(
+        build_fn(error = "crate::error::BuilderError"),
+        setter(strip_option, into)
+    )
+)]
+#[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
+#[cfg_attr(feature = "validate", derive(validator::Validate))]
+pub struct AccessRuleCreate {
+    #[cfg_attr(feature = "builder", builder(default))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[cfg_attr(feature = "validate", validate(length(min = 1, max = 64)))]
+    pub id: Option<String>,
+
+    #[cfg_attr(feature = "builder", builder(default))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[cfg_attr(feature = "validate", validate(length(min = 1, max = 16)))]
+    pub method: Option<String>,
+
+    #[cfg_attr(feature = "builder", builder(default))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[cfg_attr(feature = "validate", validate(length(min = 1, max = 128)))]
+    pub path: Option<String>,
+
+    #[cfg_attr(feature = "builder", builder(default))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[cfg_attr(feature = "validate", validate(length(min = 1, max = 64)))]
+    pub service: Option<String>,
+}
--- a/crates/api-types/src/v3/application_credential/access_rule.rs
+++ b/crates/api-types/src/v3/application_credential/access_rule.rs
@@ -14,6 +14,11 @@
 
 use serde::{Deserialize, Serialize};
 /// Short access rule representation.
+///
+/// Access rules are fine-grained permissions attached to application
+/// credentials. Each rule constrains the credential to a specific service
+/// type, HTTP method, and API path. Once created, an access rule can be
+/// viewed and deleted independently of the application credential.
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 #[cfg_attr(
     feature = "builder",
@@ -26,23 +31,30 @@ use serde::{Deserialize, Serialize};
 #[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
 #[cfg_attr(feature = "validate", derive(validator::Validate))]
 pub struct AccessRule {
-    /// The ID of the access rule.
+    /// Unique identifier of the access rule. This ID can be used to reuse an
+    /// existing access rule when creating a new application credential.
     #[cfg_attr(feature = "validate", validate(length(min = 1, max = 64)))]
     pub id: String,
 
-    /// The HTTP method permitted.
+    /// HTTP method that this access rule permits (e.g., `GET`, `POST`, `PUT`,
+    /// `DELETE`, `PATCH`).
     #[cfg_attr(feature = "builder", builder(default))]
     #[serde(skip_serializing_if = "Option::is_none")]
     #[cfg_attr(feature = "validate", validate(length(min = 1, max = 16)))]
     pub method: Option<String>,
 
-    /// The API path permitted.
+    /// API path pattern that this access rule permits. Supports wildcard
+    /// syntax: `*` matches a single path segment, `**` matches any number
+    /// of segments recursively, and `{variable}` matches a named path
+    /// parameter. For example, `/v2.1/servers/*/ips` or `/v2.1/**`.
     #[cfg_attr(feature = "builder", builder(default))]
     #[serde(skip_serializing_if = "Option::is_none")]
     #[cfg_attr(feature = "validate", validate(length(min = 1, max = 128)))]
     pub path: Option<String>,
 
-    /// The service type permitted.
+    /// OpenStack service type that this access rule applies to
+    /// (e.g., `compute`, `monitoring`, `identity`). Matched against the
+    /// service catalog.
     #[cfg_attr(feature = "builder", builder(default))]
     #[serde(skip_serializing_if = "Option::is_none")]
     #[cfg_attr(feature = "validate", validate(length(min = 1, max = 64)))]
@@ -50,6 +62,11 @@ pub struct AccessRule {
 }
 
 /// Access rule for creation (id is optional).
+///
+/// When creating an application credential, access rules can either be
+/// defined inline (with `method`, `path`, and `service`) or reference an
+/// existing rule by its `id`. All fields are optional so that both creation
+/// patterns are supported.
 #[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 #[cfg_attr(
     feature = "builder",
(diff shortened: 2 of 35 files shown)


crates/keystone/src/api/v3/user/application_credential/create.rs
```diff
@@ -0,0 +1,410 @@
...
+    State(state): State<ServiceState>,
+    Json(payload): Json<ApplicationCredentialCreateRequest>,
+) -> Result<impl IntoResponse, KeystoneApiError> {
+    payload.validate()?;
+
+    // Verify user exists — 404 if not found
+    state
+        .provider
+        .get_identity_provider()
+        .get_user(&ExecutionContext::from_auth(&state, &user_auth), &user_id)
```
gtema: you should only create ExecutionContext once. You recreate it few times in every handler

crates/api-types/src/v3/application_credential/application_credential.rs
gtema, line 1: please add docstring to all fields in structures

crates/api-types/src/v3/application_credential/access_rule.rs
gtema, line 1: please add docstring to all fields in structures

crates/api-types/src/v3/application_credential/application_credential.rs
```diff
@@ -0,0 +1,179 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expires_at: Option<DateTime<Utc>>,
+
+    pub id: String,
+    pub name: String,
+    pub project_id: String,
+    pub roles: Vec<RoleRef>,
+
+    /// Only present in create response. Never returned again.
+    pub secret: String,
```
gtema: make this SecretString

crates/api-types/src/v3/application_credential/application_credential.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    #[cfg_attr(feature = "validate", validate(length(min = 1, max = 64)))]
+    pub project_id: String,
+
+    #[cfg_attr(feature = "validate", validate(nested))]
+    pub roles: Vec<RoleRef>,
+
+    pub unrestricted: bool,
+
+    #[cfg_attr(feature = "validate", validate(length(min = 1, max = 64)))]
```
gtema: python keystone does not expose user_id for the application credential, since they are always already bound to the URL. Drop it in all structs

crates/appcred-driver-sql/src/application_credential/delete.rs
```diff
@@ -0,0 +1,99 @@
...
+use crate::entity::{
+    application_credential as db_application_credential,
+    prelude::ApplicationCredential as DbApplicationCredential,
+};
+
+pub async fn delete(
+    db: &DatabaseConnection,
+    id: &str,
+) -> Result<(), ApplicationCredentialProviderError> {
+    let app_cred = DbApplicationCredential::find()
```
gtema: let's rather only do `let res = DbApplicationCredential::delete_by_id ...  if res.rows_affected == 1 {Ok(())} else  {return Err(ApplicationCredentialProviderError::ApplicationCredentialNotFound)....` There is no need to fetch the row

crates/core/src/application_credential/service.rs
```diff
@@ -268,6 +268,48 @@ impl ApplicationCredentialApi for ApplicationCredentialService {
...
+    ///
+    /// # Parameters
+    /// - `state`: The current service state.
+    /// - `id`: The ID of the application credential to delete.
+    ///
+    /// # Returns
+    /// - `Result<(), ApplicationCredentialProviderError>` - Unit on success, or
+    ///   an error.
+    async fn delete_application_credential<'a>(
+        &self,
```
gtema, line 303: this implementation is based on the old code. Now you need to wrap it into the audited_op! macro for emitting audit record

crates/core/src/application_credential/service.rs
```diff
@@ -268,6 +268,48 @@ impl ApplicationCredentialApi for ApplicationCredentialService {
...
+        ctx: &ExecutionContext<'a>,
+        id: &'a str,
+    ) -> Result<(), ApplicationCredentialProviderError> {
+        // Fetch first to get project_id for the event — mirrors Python fetching before delete
+        let app_cred = self
```
gtema: this is not necessary, especially that with the other fetching you fetch the record 3 times before actually deleting it. Python is doing it only to be able to raise 404 what you can do even without pre-fetching

crates/keystone/src/api/v3/user/application_credential/create.rs
```diff
@@ -0,0 +1,410 @@
...
+        .await?
+        .ok_or_else(|| KeystoneApiError::not_found("user", &user_id))?;
+
+    // Security check — cannot create credentials for another user
+    let ctx_user_id = user_auth.principal().get_user_id();
```
gtema: the check is very good, but it would be good to place it before verifying the user exist. This helps preventing timing attacks.

crates/keystone/src/api/v3/user/application_credential/create.rs
```diff
@@ -0,0 +1,410 @@
...
+            ));
+        }
+    };
+
+    state
+        .policy_enforcer
+        .enforce(
+            "identity/user/application_credential/create",
+            &user_auth,
+            json!({"user_id": user_id}),
```
gtema: We actually need to pass the whole ApplicationCredential struct into the policy, not only user_id

crates/api-types/src/v3/application_credential/application_credential.rs
```diff
@@ -0,0 +1,179 @@
...
+    feature = "builder",
+    derive(derive_builder::Builder),
+    builder(
+        build_fn(error = "crate::error::BuilderError"),
+        setter(strip_option, into)
+    )
+)]
+#[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
+#[cfg_attr(feature = "validate", derive(validator::Validate))]
+pub struct ApplicationCredentialCreate {
```
gtema, line 96: python keystone allows to specify (optionally) the secret during creation. Add it wrapped as SecretString

crates/api-types/src/v3/application_credential_conv.rs
```diff
@@ -0,0 +1,117 @@
...
+    fn from(value: api_types_application_credential::ApplicationCredentialCreate) -> Self {
+        Self {
+            access_rules: value
+                .access_rules
+                .map(|rules| rules.into_iter().map(Into::into).collect()),
+            description: value.description,
+            expires_at: value.expires_at,
+            id: value.id,
+            name: value.name,
+            project_id: String::new(), // assigned server-side from token
```
gtema: this is pretty bad, you should not set it as empty string

crates/api-types/src/v3/application_credential_conv.rs
```diff
@@ -0,0 +1,117 @@
...
+            roles: value.roles.into_iter().map(Into::into).collect(),
+            secret: None, // generated server-side
+            unrestricted: value.unrestricted,
+            user_id: String::new(), // assigned server-side from token
```
gtema: also here - no empty strings

crates/api-types/src/v3/application_credential_conv.rs
```diff
@@ -0,0 +1,117 @@
...
+}
+
+impl From<api_types_access_rule::AccessRuleCreate> for provider_types::AccessRuleCreate {
+    fn from(value: api_types_access_rule::AccessRuleCreate) -> Self {
+        Self {
+            id: value.id,
+            method: value.method,
+            path: value.path,
+            service: value.service,
+            user_id: String::new(), // assigned server-side
```
gtema: no empty strings

crates/keystone/src/api/v3/user/application_credential/create.rs
```diff
@@ -0,0 +1,410 @@
...
+            None,
+        )
+        .await?;
+
+    let mut app_cred: openstack_keystone_core_types::application_credential::ApplicationCredentialCreate
```
gtema: ApplicationCredentialCreate in the core type supports builder context. This is how you can pre-populate all regular fields (without project_id and user_id). Then set 2 remaining fields and build the builder into the core request.

crates/keystone/src/api/v3/user/application_credential/create.rs
```diff
@@ -0,0 +1,410 @@
...
+    let project_id = match user_auth.authorization().map(|a| &a.scope) {
+        Some(ScopeInfo::Project { project, .. }) => project.id.clone(),
+        _ => {
+            return Err(KeystoneApiError::BadRequest(
+                "application credentials require a project-scoped token".into(),
+            ));
+        }
+    };
+
+    state
```
gtema, line 77: I would suggest you should swap order like that:
- call policy passing the application credentials with user_id (dump the struct to json, inject user_id and pass it as "application_credential")
- the policy evaluates wheter input.credential.user_id == target.application_credential.user_id
- check user existence

crates/keystone/src/api/v3/user/application_credential/delete.rs
```diff
@@ -0,0 +1,302 @@
...
+        (status = 401, description = "Unauthorized"),
+    ),
+    tag = "application_credentials"
+)]
+pub(super) async fn delete(
+    Auth(user_auth): Auth,
+    Path((user_id, application_credential_id)): Path<(String, String)>,
+    State(state): State<ServiceState>,
+) -> Result<impl IntoResponse, KeystoneApiError> {
+    state
```
gtema: this should not be done here at all

crates/keystone/src/api/v3/user/application_credential/delete.rs
```diff
@@ -0,0 +1,302 @@
...
+            KeystoneApiError::not_found("application_credential", &application_credential_id)
+        })?;
+
+    // Use credential's real user_id — not the URL parameter
+    state
+        .policy_enforcer
+        .enforce(
+            "identity/user/application_credential/delete",
+            &user_auth,
+            json!({"user_id": current.user_id}),
```
gtema: you need to pass proper application credential into the policy - read ADR 0002. First param is emty and the second is `current`

crates/keystone/src/api/v3/user/application_credential/list.rs
```diff
@@ -0,0 +1,324 @@
...
+        .get_user(&ExecutionContext::from_auth(&state, &user_auth), &user_id)
+        .await?
+        .ok_or_else(|| KeystoneApiError::not_found("user", &user_id))?;
+
+    state
+        .policy_enforcer
+        .enforce(
+            "identity/user/application_credential/list",
+            &user_auth,
+            json!({"user_id": user_id}),
```
gtema: every time you write http handler double check the ADR 0002 - you need to always pass data into the policy. It also dictates order of calls

crates/keystone/src/api/v3/user/application_credential/list.rs
```diff
@@ -0,0 +1,324 @@
...
+pub(super) async fn list(
+    Auth(user_auth): Auth,
+    Path(user_id): Path<String>,
+    Query(payload): Query<ApplicationCredentialListParameters>,
+    State(state): State<ServiceState>,
+) -> Result<impl IntoResponse, KeystoneApiError> {
+    payload.validate()?;
+
+    // Verify user exists — returns 404 if not found per OpenStack API spec
+    state
```
gtema: must not happen here

crates/keystone/src/api/v3/user/application_credential/show.rs
```diff
@@ -0,0 +1,316 @@
...
+        .ok_or_else(|| {
+            KeystoneApiError::not_found("application_credential", &application_credential_id)
+        })?;
+
+    state
+        .policy_enforcer
+        .enforce(
+            "identity/user/application_credential/show",
+            &user_auth,
+            json!({"user_id": current.user_id}),
```
gtema: same policy rules as mentioned above

policy/user/application_credential/create.rego
```diff
@@ -0,0 +1,9 @@
+# METADATA
+# description: Policy for creating application credentials
+package identity.user.application_credential.create
+
+default allow := false
+
+allow if {
+    input.credentials.user_id == input.target.user_id
+}
```
gtema, line 13: please also add a "violation" population that adds a context when credentials.user_id != target.application_credential.user_id

policy/user/application_credential/create.rego
```diff
@@ -0,0 +1,9 @@
+# METADATA
+# description: Policy for creating application credentials
+package identity.user.application_credential.create
+
+default allow := false
+
+allow if {
+    input.credentials.user_id == input.target.user_id
```
gtema: it would be input.target.application_credential.user_id

policy/user/application_credential/delete.rego
```diff
@@ -0,0 +1,9 @@
+# METADATA
+# description: Policy for deleting application credentials
+package identity.user.application_credential.delete
+
+default allow := false
+
+allow if { input.credentials.is_admin }
+
+allow if { input.credentials.user_id == input.target.user_id }
```
gtema: it's going to be input.target.existing.user_id

policy/user/application_credential/list.rego
```diff
@@ -0,0 +1,14 @@
...
+default allow := false
+
+allow if { input.credentials.is_admin }
+
+allow if {
+    "reader" in input.credentials.roles
+    input.credentials.system == "all"
+}
+
+allow if { input.credentials.user_id == input.target.user_id }
```
gtema: target.application_credential.user_id or target.user_id, but depends on how you set pass it from handler. Anyway - please look at the current policies which document the structure of the data - add it to all policies

policy/user/application_credential/show.rego
```diff
@@ -0,0 +1,14 @@
...
+default allow := false
+
+allow if { input.credentials.is_admin }
+
+allow if {
+    "reader" in input.credentials.roles
+    input.credentials.system == "all"
+}
+
+allow if { input.credentials.user_id == input.target.user_id }
```
gtema: also here user_id is wrapped under the input.existing.application_credential.user_id

tests/api/tests/api_v3/identity/application_credential/list.rs
```diff
@@ -0,0 +1,88 @@
...
+use test_api::guard::ResourceGuard;
+use test_api::identity::application_credential::{
+    create_application_credential, list_application_credentials,
+};
+use tracing_test::traced_test;
+
+pub async fn get_project_scoped_client() -> Result<Arc<AsyncOpenStack>> {
+    let mut tc = AsyncOpenStack::new(&CloudConfig::from_env()?).await?;
+
+    tc.authorize(
```
gtema, line 14: this should not be necessary - the 'tc' already should be project scoped as directed by the env vars.

gtema (changes requested):
thanks, look good, but there are issues that need to be addressed

konac-hamza:
> thanks, look good, but there are issues that need to be addressed

Thanks for your review. I'm going to address all issues.

crates/keystone/src/api/v3/user/application_credential/create.rs
konac-hamza: I will define it at begin of the handler and reuse it in the handler if I need to use it.

crates/api-types/src/v3/application_credential/application_credential.rs
konac-hamza: okey I will add by the help of AI

crates/core/src/application_credential/service.rs
konac-hamza: Instead of refetch it, I will update service method to get ApplicationCredential struct as paramater.

crates/api-types/src/error_conv.rs
```diff
@@ -239,19 +239,33 @@ impl From<CatalogProviderError> for KeystoneApiError {
...
-            err @ ApplicationCredentialProviderError::ApplicationCredentialExpired => {
-                Self::unauthorized(err, None::<String>)
+            ref err @ ApplicationCredentialProviderError::Validation { .. } => {
+                Self::BadRequest(err.to_string())
+            }
+            ApplicationCredentialProviderError::ApplicationCredentialExpired => {
+                Self::BadRequest("application credential has expired".into())
+            }
+            ApplicationCredentialProviderError::AccessRuleInUse(x) => {
+                Self::BadRequest(format!("access rule {x} is still in use"))
```
gtema: InUse is a typical 409 - conflict and not bad request. User 'Self::Conflict' instead

crates/api-types/src/v3/application_credential/application_credential.rs
```diff
@@ -0,0 +1,228 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub description: Option<String>,
+
+    /// Optional expiration date and time. When `None`, the credential does not
+    /// expire.
+    #[cfg_attr(feature = "builder", builder(default))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expires_at: Option<DateTime<Utc>>,
+
+    /// Optional client-supplied identifier for the application credential. If
```
gtema: Passing ID in create through API must not be possible - inconsistency to python Keystone

tests/api/tests/api_v3/identity/application_credential/create.rs
```diff
@@ -0,0 +1,76 @@
...
+        ApplicationCredentialCreateBuilder::default()
+            .name("test-cred")
+            .roles(vec![])
+            .build()?,
+    )
+    .await?;
+
+    assert_eq!(cred.name, "test-cred");
+    assert!(!cred.secret.is_empty());
+    // assert_eq!(cred.user_id, user_id);
```
gtema: why commented out?

tests/api/tests/api_v3/identity/application_credential/list.rs
```diff
@@ -0,0 +1,71 @@
...
+        .ok_or_else(|| eyre::eyre!("no auth info available"))?
+        .token
+        .user
+        .id;
+
+    let cred1 = create_application_credential(
+        &tc,
+        &user_id,
+        ApplicationCredentialCreateBuilder::default()
+            .name(format!("cred-1"))
```
gtema: just use `.name("cred-1")` - Builder should already have Into

crates/keystone/src/api/v3/domain/mod.rs
```diff
@@ -27,7 +27,8 @@ pub mod types;
 #[derive(OpenApi)]
 #[openapi(
     tags(
-        (name="domains", description=r#"Domains are a collection of projects and users that define administrative boundaries for managing Identity entities. Domains can represent an individual, company, or operator-owned space. They expose administrative activities directly to system users. Users can be granted the administrator role for a domain. A domain administrator can create projects, users, and groups in a domain and assign roles to users and groups in a domain.
```
gtema, line 31: not related change - please undo

gtema (changes requested):
Very very close, just few nits

tests/api/tests/api_v3/identity/application_credential/create.rs
konac-hamza: it is mistake, i've forgotten to uncomment after debugging

crates/keystone/src/api/v3/domain/mod.rs
konac-hamza: thanks for review, I have no idea when I changed this file.I will be more careful after that 👍

konac-hamza:
@gtema I didnt understand  reason of failure in the CI tests.The /health endpoint doesn't response respect of the logs. I have checked last merged commits but none of them has issue like this,  Should I change anything in the PR?

crates/api-types/src/v3/application_credential/application_credential.rs
gtema: this is still open

crates/api-types/src/error_conv.rs
```diff
@@ -233,25 +233,51 @@
...
             }
-            err @ ApplicationCredentialProviderError::ApplicationCredentialExpired => {
-                Self::unauthorized(err, None::<String>)
+            ref err @ ApplicationCredentialProviderError::Validation { .. } => {
+                Self::BadRequest(err.to_string())
+            }
+            ApplicationCredentialProviderError::ApplicationCredentialExpired => {
+                Self::BadRequest("application credential has expired".into())
+            }
+            ApplicationCredentialProviderError::AccessRuleInUse(x) => {
```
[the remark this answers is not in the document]
gtema: do `ApplicationCredentialProviderError::AccessRuleInUse(_)` to keep clippy happy

crates/keystone/src/api/v3/user/application_credential/delete.rs
```diff
@@ -0,0 +1,295 @@
...
+            KeystoneApiError::not_found("application_credential", &application_credential_id)
+        })?;
+
+    // Policy check — uses credential's real user_id, not URL parameter
+    state
+        .policy_enforcer
+        .enforce(
+            "identity/user/application_credential/delete",
+            &user_auth,
+            json!({"application_credential": serde_json::to_value(&current)?}),
```
gtema: wrong order: policy_name, auth, target, current. You pass current under target

crates/keystone/src/api/v3/user/application_credential/show.rs
```diff
@@ -0,0 +1,311 @@
...
+            KeystoneApiError::not_found("application_credential", &application_credential_id)
+        })?;
+
+    // Policy check — uses stored object's real user_id
+    state
+        .policy_enforcer
+        .enforce(
+            "identity/user/application_credential/show",
+            &user_auth,
+            json!({"application_credential": serde_json::to_value(&current)?}),
```
gtema: also here wrong order of target/current

policy/user/application_credential/list.rego
```diff
@@ -0,0 +1,20 @@
...
+default allow := false
+
+allow if { input.credentials.is_admin }
+
+allow if {
+    "reader" in input.credentials.roles
+    input.credentials.system == "all"
+}
+
+allow if { input.credentials.user_id == input.target.application_credential.user_id }
```
gtema, line 14: the check in this form tells that while input.target.application_credential.user_id = none would not be allowed, we should throw dedicated violation message (smth like listing application credentials require user_id ...)

policy/user/application_credential/delete.rego
```diff
@@ -0,0 +1,14 @@
+# METADATA
+# description: Policy for deleting application credentials
+package identity.user.application_credential.delete
+
+default allow := false
+
+allow if { input.credentials.is_admin }
+
+allow if { input.credentials.user_id == input.target.application_credential.user_id }
```
gtema: here it matches the handler, but not the general policy: delete/update/show take input.current and not input.target

gtema (changes requested):
good progress, there are still some issues. I am currently working on adding more jobs/gates to prevent some of the issues you have here in policies

crates/keystone/src/api/v3/user/application_credential/delete.rs
konac-hamza: - `- **Show**: Pass `serde_json::to_value(stored_object)?` as target, `None` as
  existing ` 
- `- **Delete**: Pass `serde_json::to_value(stored_object)?` as target, `None` as
  existing`
  
  Based on ADR doc the current object must pass as target. Should I still change it?

gtema: nice spot - this is a mistake in the ADR. Everywhere in the code we follow this convention (logical that "existing" parameter always points at the object stored on the server side). So yes, please change it and I fix the ADR

crates/api-types/src/v3/application_credential/application_credential.rs
gtema: you still do not have a "secret" specified by the user during AC creation.

crates/api-types/src/v3/application_credential_conv.rs
```diff
@@ -0,0 +1,144 @@
...
+            name: value.name,
+            project_id: value.project_id,
+            roles: value.roles.into_iter().map(Into::into).collect(),
+            secret: value.secret.expose_secret().to_string().into(),
+            unrestricted: value.unrestricted,
+        }
+    }
+}
+
+impl From<api_types_application_credential::ApplicationCredentialCreate>
```
gtema, line 71: when user passes the expected secret we would need to set it here as well

crates/keystone/src/api/v3/user/application_credential/delete.rs
```diff
@@ -0,0 +1,295 @@
...
+            KeystoneApiError::not_found("application_credential", &application_credential_id)
+        })?;
+
+    // Policy check — uses credential's real user_id, not URL parameter
+    state
+        .policy_enforcer
+        .enforce(
+            "identity/user/application_credential/delete",
+            &user_auth,
+            json!({"application_credential": null}),
```
gtema: no, simply `serde_json::Value::Null`

crates/keystone/src/api/v3/user/application_credential/create.rs
```diff
@@ -0,0 +1,397 @@
...
+        Some(ScopeInfo::Project { project, .. }) => project.id.clone(),
+        _ => {
+            return Err(KeystoneApiError::BadRequest(
+                "application credentials require a project-scoped token".into(),
+            ));
+        }
+    };
+
+    let mut target = serde_json::to_value(&payload.application_credential)?;
+    target["user_id"] = json!(user_id);
```
gtema, line 70: once you add "secret" to the payload you need to explicitly remove it from the json here so that it is not passed to the policy

crates/keystone/src/api/v3/user/application_credential/show.rs
```diff
@@ -0,0 +1,311 @@
...
+            KeystoneApiError::not_found("application_credential", &application_credential_id)
+        })?;
+
+    // Policy check — uses stored object's real user_id
+    state
+        .policy_enforcer
+        .enforce(
+            "identity/user/application_credential/show",
+            &user_auth,
+            json!({"application_credential": null}),
```
gtema: simply use `serde_json::Value::Null`

gtema (changes requested):
few more remaining issues, very close now

crates/api-types/src/v3/application_credential/application_credential.rs
```diff
@@ -0,0 +1,256 @@
...
+#[cfg_attr(feature = "openapi", derive(utoipa::IntoParams))]
+#[cfg_attr(feature = "validate", derive(validator::Validate))]
+pub struct ApplicationCredentialListParameters {
+    /// Optional name filter. When set, only application credentials whose
+    /// name matches this value are returned.
+    #[cfg_attr(feature = "validate", validate(length(max = 255)))]
+    pub name: Option<String>,
+}
+
+fn serialize_secret<S>(secret: &SecretString, serializer: S) -> Result<S::Ok, S::Error>
```
gtema: you do not need to introduce serialize helpers, they are present in api-types/src/common.rs

crates/api-types/src/v3/application_credential_conv.rs
```diff
@@ -0,0 +1,159 @@
...
+            access_rules: value
+                .access_rules
+                .map(|rules| rules.into_iter().map(Into::into).collect()),
+            description: value.description,
+            expires_at: value.expires_at,
+            id: value.id,
+            name: value.name,
+            project_id: value.project_id,
+            roles: value.roles.into_iter().map(Into::into).collect(),
+            secret: value.secret.expose_secret().to_string().into(),
```
gtema, line 65: why are you exposing secret to put it into the secret?

crates/keystone/src/api/v3/user/application_credential/create.rs
```diff
@@ -0,0 +1,400 @@
...
+    // Check payload has secret field,if yes copy target variable and remove secret from copied variable and pass it to policy enforcer
+    let mut target_for_policy = target.clone();
+    target_for_policy.as_object_mut().unwrap().remove("secret");
```
gtema: appparently clippy job is not running properly - we do not use unwrap/expect in the production code.

gtema (changes requested):
small nits to fix

crates/api-types/src/v3/application_credential_conv.rs
konac-hamza: You're right — since both sides are SecretString, I should pass it directly.

crates/keystone/src/api/v3/user/application_credential/create.rs
```diff
@@ -0,0 +1,402 @@
...
+        )
+        .await?;
+
+    let app_cred = core_type_application_credential::ApplicationCredentialCreateBuilder::from(
+        payload.application_credential,
+    )
+    .user_id(user_id.clone())
+    .project_id(project_id)
+    .build()
+    .unwrap();
```
gtema: cargo clippy forbids use of unwrap in prod code, setup was recently fixed to apply global rules to the main crate as well and since you pushed not from the up2date branch it may be still not running. Run `cargo clippy` on the workspace locally to ensure this

konac-hamza: Yes, my workspace gives error because of unwrap usage. So I will rebase the branch and update the PR.

tests/api/src/identity/application_credential.rs
```diff
@@ -0,0 +1,188 @@
...
+use openstack_sdk::{AsyncOpenStack, api::QueryAsync};
+use std::borrow::Cow;
+use std::sync::Arc;
+
+struct AppCredCreateRequest {
+    user_id: String,
+    app_cred: ApplicationCredentialCreate,
+}
+
+impl RestEndpoint for AppCredCreateRequest {
```
gtema, line 27: in the meanwhile the macro helpers landed to reduce amount of hand written code (see https://github.com/openstack-experimental/keystone/blob/main/tests/api/src/trust.rs for example)

gtema (changes requested):
test failure is real but caused by the bug in the entity definition: python keystone defines a unique constraint for application credential over user_id+name, so in the api tests you need to use same randomness for the appcred name. It is not failing in the regular integration tests and api test outside of k3s since only in k3s db schema is initialized by python keystone with this constraint

I am working on the fix to the entity so that it does not interfere with your change, but you must still use unique appcred names in the tests

gtema:
#1150 is the PR fixing appcred uniqueness

tests/api/src/identity/application_credential.rs
konac-hamza: i have looked at the referenced file and other samples together but I see the macro doesn't support sub-resources implementation. However, in other PR I will use this macro. Thanks for the review and recommendation.

macro-doc:
``` 
Endpoints that do not
//! fit these shapes (sub-resources, grants, borrowed fields, non-JSON
//! bodies) should keep hand-written `RestEndpoint` impls.
```

gtema (approved):
thanks

PR #996, opened 2026-07-10
feat(identity): Add per-user auth rate limiting (ADR-0022 phase 2)

### LINKED ISSUE #843: Implement rate limiting

magdang:
## Summary

- Implements ADR-0022 phase 2: the per-user `[rate_limit_user_auth]` bucket that phase 1 (#846) deferred.
- Adds a `governor`-backed per-user limiter keyed on the canonical database user ID, wired into both password and TOTP authentication.
- Fires the per-user check **after** the user is confirmed to exist and **before** any lockout read, failed-auth write, or password hashing (ADR-0022 Invariants 4 and 8).
- Maps rejections to the unified `429 Too Many Requests` with `Retry-After` already shared by the IP and API Key limiters (Invariant 3).

> Phase 1 (#846) is now merged into `main`, so this PR branches off latest `main` and targets it directly. It reuses the `RateLimitState` framework from phase 1 and adds the per-user bucket as a new field, exactly the "one-field addition" the phase 1 framework was built for.

## ADR-0022 invariants addressed

| # | Invariant | How |
|---|-----------|-----|
| 7 | Username normalization | Keyed on the backend user ID, not any user-supplied string. There is no input-derived key to normalize, and homoglyph usernames cannot split one account across buckets. Satisfied by construction, with no normalization dependency added. |
| 8 | Post-lookup per-user throttle | The check runs in the SQL driver's `authenticate_by_password` only after the user row is confirmed, and before hashing. Novel or made-up usernames never reach the limiter, so they cannot mint fresh buckets, and rejected attempts cost no CPU. |

Invariants 1 through 6 were established in phase 1 and are reused unchanged. The per-user bucket is a distinct `Arc<DefaultKeyedRateLimiter<String>>` (Invariant 5) and inherits the same fail-hard config validation (Invariant 2) and monotonic clock (Invariant 6).

## Shared bucket across password and TOTP

The TOTP path in the core service shares the same per-user bucket as the password path. This stops a client from doubling its per-user quota by alternating between the two methods. Before this change the TOTP path had no brute-force control at all. The TOTP check fires before passcode verification.

## Rejection handling

A rejection becomes a new `IdentityProviderError::TooManyRequests`, which flows through the existing `TooManyRequests { retry_after }` mapping to a 429 body plus `Retry-After` header (Invariant 3). It does **not** feed the ADR-0010 lockout counter, so a rate-limited attempt is not recorded as a failed login.

## Question for maintainer

The password check lives inside the SQL driver, which couples the persistence layer to `RateLimitState`. I chose this because it is the only point where user existence is known before hashing. A reviewer might prefer a two-phase backend trait instead (look up the user, then verify) so the limiter can live in the service layer. Happy to reshape it if that is preferred.

## Files changed

| File | Change |
|------|--------|
| `crates/config/src/lib.rs` | Add `rate_limit_user_auth: RateLimitSection` to `Config` |
| `crates/core-types/src/identity/error.rs` | Add `IdentityProviderError::TooManyRequests` |
| `crates/core/src/rate_limit.rs` | Add `check_user`, per-user bucket build, unit tests |
| `crates/core/src/identity/service.rs` | Per-user check on the TOTP path, before passcode verify |
| `crates/identity-driver-sql/src/authenticate.rs` | Per-user check in `authenticate_by_password`, post-lookup pre-hash, tests |
| `crates/identity-driver-sql/src/lib.rs` | Thread `RateLimitState` into the driver |
| `crates/api-types/src/error_conv.rs` | Map the new provider error to the unified 429 + `Retry-After` |
| `crates/keystone/src/api/v3/auth/token/create.rs` | Handler-level 429 surface test |
| `tools/keystone.conf` | Document `[rate_limit_user_auth]` with default values |

## Test plan

- [x] `cargo fmt` clean
- [x] `cargo clippy --lib --tests --workspace` clean on the changed files
- [x] Unit tests pass across `core` / `identity-driver-sql` / `api-types` / `keystone` (122 passing in core plus the driver, api-types, and keystone suites, 0 failed)
- [x] Per-user bucket is independent from the IP limiter
- [x] Hot reload of the limiter config
- [x] Rejection path runs exactly one query and no more
- [x] Nonexistent users never touch the limiter store
- [x] TOTP rejection happens before passcode verification
- [x] Handler returns a 429 to the client

## Out of scope (follow-ups)

- The 10k-entry LRU store cap from the ADR. Phase 1 shipped only the 60-second `retain_recent` trim, and governor's default store cannot do LRU without a custom state store, so it is left for a follow-up.
- Rate limiting on the federation and application-credential endpoints, which the ADR punts to a separate follow-up ADR.

Because of those two items this PR **addresses** rather than closes #843. If the maintainer considers them in scope for #843, please leave the issue open after merge.

### DIFF
--- a/crates/identity-driver-sql/src/user/get.rs
+++ b/crates/identity-driver-sql/src/user/get.rs
@@ -17,13 +17,14 @@ use sea_orm::entity::*;
 use sea_orm::query::*;
 
 use openstack_keystone_config::Config;
+use openstack_keystone_core::auth::AuthenticationError;
 use openstack_keystone_core::error::DbContextExt;
 use openstack_keystone_core::identity::IdentityProviderError;
 use openstack_keystone_core_types::identity::{UserOptions, UserResponse, UserResponseBuilder};
 
 use crate::entity::{
-    nonlocal_user as db_nonlocal_user,
-    prelude::{FederatedUser, NonlocalUser, User as DbUser, UserOption},
+    local_user as db_local_user, nonlocal_user as db_nonlocal_user,
+    prelude::{FederatedUser, LocalUser, NonlocalUser, User as DbUser, UserOption},
     user as db_user,
 };
 use crate::federated_user::MergeFederatedUserData;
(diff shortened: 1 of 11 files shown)


crates/identity-driver-sql/src/authenticate.rs
gtema, line 1: I don't think that passing rate limiting down to the driver is a good idea, it would be required to duplicate the logic in every single driver

gtema (changes requested):
I understand why rate limiting for the password auth was passed under the driver, but we need to implement it differently, otherwise every single driver need to repeat it what is not good. auth by totp checks user existence on the provider level. I fear we need to do the same for the password auth by adding new backend method for lookup of the user existence without doing expensive lookups (e.g., check_user_exist). This method can only check presence of the main db table entry with no joins and only return `Result<String, IdentityProviderError>` (checking user for being disabled) to allow resolution by user_id or name and deduplicating logic part for password and totp auth

crates/identity-driver-sql/src/authenticate.rs
magdang: Agreed, and implemented as you described.

Added `IdentityBackend::check_user_exist`:

```rust
async fn check_user_exist<'a>(
    &self,
    state: &ServiceState,
    user_id: Option<&'a str>,
    name: Option<&'a str>,
    domain_id: Option<&'a str>,
) -> Result<String, IdentityProviderError>;
```

- **Cheap by contract:** point lookups only — name resolves via `local_user`
  (exact match), falling back to `nonlocal_user`, then the main `user` entry
  for the disabled check. No joins, no password or option loading.
- **Provider-level throttle:** `IdentityService` now resolves the reference and
  applies the `[rate_limit_user_auth]` bucket for **both** password and TOTP,
  keyed on the returned canonical ID. The TOTP path reuses the same probe
  instead of its own `get_user`/`list_users` resolution, so the lookup logic is
  deduplicated as well.
- **Driver decoupled:** `authenticate.rs` in the SQL driver is back to its
  pre-PR shape; drivers only implement the probe.

Three details worth noting:

1. A probe `UserNotFound` falls through to the backend, so the dummy-hash
   timing path and the uniform 401 are unchanged — unknown names still never
   touch the limiter store (Invariant 8).
2. On the password path the probe only runs when `[rate_limit_user_auth]` is
   enabled, so the default hot path gains no query. TOTP always uses it, since
   it replaced the existing resolution.
3. Behavior nuance: with the limiter enabled, a disabled user is now rejected
   by the probe **before** the ADR-0010 lockout check evaluates (previously
   lockout was checked first). Both cases return 401 either way ;  happy to
   adjust if that matters.

magdang: I also moved out of the driver. The throttle now lives in
`IdentityService::authenticate_by_password`, keyed on the ID returned by the
new `check_user_exist` probe, and this file is back to its pre-PR state. The
driver-level tests moved to the service level accordingly.

crates/identity-driver-sql/src/user/get.rs
```diff
@@ -126,6 +127,69 @@ pub async fn get(
...
+/// - `db`: The database connection.
+/// - `user_id`: The user ID, when resolving by ID.
+/// - `name`: The user name, when resolving by name.
+/// - `domain_id`: The domain ID owning `name`.
+///
+/// # Returns
+/// A `Result` containing the canonical user ID, or an `Error`
+/// (`UserNotFound` when absent, `UserDisabled` when disabled).
+#[tracing::instrument(skip_all)]
+pub async fn check_user_exist(
```
gtema, line 149: in the password/totp auth this is the correct approach to only look for the local user, for the federated/nonlocal users it is not. I mean that in general when the function is invoked from password auth flow it works correctly, but it will not find federated users should someone want this. Let's fix this in the followup not to disturb the change context too much

gtema (approved):

PR #1022, opened 2026-07-15
test: include request ID in API authentication errors

### LINKED ISSUE #999: Extend api test common suite to capture request-id
When api tests fail (in k8s or locally) and the only thing the user get's back is `[api-test-v3] Error: Authentication failed with 401 Unauthorized: {"error":{"code":401,"message":"The request you have made requires authentication."}}` is not very helpful. It would be great when also request id, which is returned in the headers is also returned with the error message to ease searching in the logs for the corresponding transaction.

MohidSheraz:
## Summary

Include the `x-openstack-request-id` response header in authentication errors produced by the shared API test client.

The authentication error helper now preserves:

- HTTP status
- request ID, when present
- response body, when present

The helper is used by the common authentication, rescope, and password-authentication paths.

Closes #999.

## Testing

- Ran Rust formatting in Docker
- Verified the diff with `git diff --check`
- Built the Skaffold Docker image, including the `integration_api_v3` test binary

### DIFF
--- a/tests/api/src/common.rs
+++ b/tests/api/src/common.rs
@@ -16,7 +16,7 @@
 use eyre::{OptionExt, Result, WrapErr, eyre};
 use openstack_sdk::{AsyncOpenStack, config::CloudConfig};
 use reqwest::{
-    Client, ClientBuilder, StatusCode,
+    Client, ClientBuilder, Response, StatusCode,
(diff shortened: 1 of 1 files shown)


gtema (approved):
This is only addressing authentication errors leaving others naked, but as PR mentions it only targets authentication errors - good to go

PR #1048, opened 2026-07-17
fix: Use system field in OPA policies

### LINKED ISSUE #1045: OPA policies check credentials.system_scope but the Rust field is `system` — system-reader rule is dead
**Found while investigating tempest-identity-compatibility 403s (#997/#998), not the direct cause of any currently-tracked failure — a separate correctness bug surfaced as a byproduct.**

`Credentials` (`crates/core/src/policy.rs`) serializes its system-scope field as `system` (`pub system: Option<String>`, no `#[serde(rename)]`), populated as `"all"` for system scope by `Credentials::try_from`. Several `.rego` policies instead check a field named `system_scope`, which is never present in the real OPA input — so their "`reader` role with system scope may read this" rule is silently dead (always evaluates to `undefined`/false), not just for these three, for every request that should qualify:

- `policy/auth/token/show.rego` (lines ~40-41)
- `policy/auth/project/list.rego`
- `policy/assignment/list.rego`
- `policy/service/show.rego`
- `policy/service/list.rego`
- `policy/endpoint/show.rego`
- `policy/endpoint/list.rego`

(`policy/auth/token/revoke.rego` has the same wrong field name but in an already-commented-out block, so it's inert either way.)

Each of these files' own `_test.rego` mocks the same wrong `system_scope` key, so the existing unit tests pass while asserting a self-consistently wrong contract — they don't catch this because they never exercise the real Rust→JSON field name.

**Impact**: a caller with `role:reader` and system scope (`GET` a token, list domains-a-project-belongs-to/projects, list role assignments, show/list services, show/list endpoints) is denied even though the policy's own doc comments say they should be allowed. Fails closed (denies access that should be granted), not a privilege escalation, but a real functional/compatibility gap — likely part of why some tempest identity tests using system-scoped reader credentials fail.

**Fix**: rename `system_scope` → `system` in each `allow` rule listed above and in the corresponding `_test.rego` mocks (matching the working reference implementation already used correctly elsewhere, e.g. `policy/resource/domain/show.rego`'s `input.credentials.system == "all"`).

MohidSheraz:
Update system-scoped reader policies to useinput.credentials.system, matching the field serialized by the RustCredentials type.

Update the corresponding Rego test inputs to use the same system field.

This restores policy rules that were previously checking the nonexistentcredentials.system_scope field.

Closes #1045.

Testing

Ran OPA formatting validation with OPA 1.16.2 in Docker

Ran OPA policy tests with OPA 1.16.2 in Docker

Ran git diff --check

Note: was made with assistance of CodeX.

### DIFF
--- a/policy/assignment/list.rego
+++ b/policy/assignment/list.rego
@@ -31,8 +31,8 @@ allow if {
 # description: "'reader' in the system scope can list any role assignments."
 allow if {
 	"reader" in input.credentials.roles
-	input.credentials.system_scope != null
-	"all" == input.credentials.system_scope
+	input.credentials.system != null
+	"all" == input.credentials.system
 }
 
 # METADATA
--- a/policy/assignment/list_test.rego
+++ b/policy/assignment/list_test.rego
@@ -4,7 +4,7 @@ import data.identity.assignment.list
 
 test_allowed if {
 	list.allow with input as {"credentials": {"roles": ["admin"]}}
-	list.allow with input as {"credentials": {"roles": ["reader"], "system_scope": "all"}}
+	list.allow with input as {"credentials": {"roles": ["reader"], "system": "all"}}
 	list.allow with input as {
 		"credentials": {"roles": ["admin"], "domain_id": "domain_a"},
 		"target": {"assignment": {"domain_id": "domain_b"}},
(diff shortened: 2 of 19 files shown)


gtema:
as expected - there is a merge conflict now

gtema (changes requested):
I assume since the PR was quite some time in the pipe few other entries appeared (or they were initially missed): policy/region/{list,show}.rego still use "system_scope". Please address them here as well. Otherwise good to go

gtema (approved):

PR #1050, opened 2026-07-17
fix: use PATCH for v3 update operations

MohidSheraz:
Summary

Updates the Identity v3 user, service, and endpoint update APIs to use PATCH instead of PUT, matching the OpenStack Identity v3 API specification.

Changes

Changed PUT to PATCH for:

PATCH /v3/users/{user_id}

PATCH /v3/services/{service_id}

PATCH /v3/endpoints/{endpoint_id}

Updated the related handler tests to send PATCH

Updated the API test client helpers to use PATCH

Left credential updates unchanged because they already use PATCH

Left v3 role grants and role implication operations as PUT because they create associations rather than update resources

Left v4 update routes unchanged; v4 support for both PUT and PATCH can be handled separately

Testing

Verified there are no remaining PUT methods in the affected v3 update handlers

Ran git diff --check

Ran formatting checks

Ran the relevant update tests

Note: used CodeX for assistance

### DIFF
--- a/crates/keystone/src/api/v3/endpoint/update.rs
+++ b/crates/keystone/src/api/v3/endpoint/update.rs
@@ -29,7 +29,7 @@ use openstack_keystone_core::auth::ExecutionContext;
 
 /// Update existing endpoint
 #[utoipa::path(
-    put,
+    patch,
     path = "/{endpoint_id}",
     description = "Update endpoint by ID",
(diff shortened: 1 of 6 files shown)


gtema (approved):
this PR uncovers that we actually lack lot of update operations, but that's a different thing

PR #1055, opened 2026-07-17
refactor: Extract password hashing into own crate

### LINKED ISSUE #946: Evaluate moving password hashing implementors out of the core crate
Currently password hashing is implemented in the openstack-keystone-core crate. This pulls additional dependencies to pretty much every crate which should not be required. Maybe it is better to keep the trait defined in the core crate, but move the hashers into the separate crate that can be compiled separately

magdang:
## TL;DR

Password hashing was implemented inside `openstack-keystone-core`, which forced
**every** crate depending on core to compile the `bcrypt` and `scrypt` crates —
even though only the identity and appcred SQL drivers (and the keystone binary's
cache reset) ever touch password hashing. This extracts the `password_hashing`
module into a new `openstack-keystone-password-hashing` crate so those crypto
crates are compiled only where they are actually used. `core` no longer pulls
`bcrypt` or `scrypt` into its dependency tree.

Closes #946.

## Changes done

- **New crate `crates/password-hashing`** (`openstack-keystone-password-hashing`):
  the whole `password_hashing` module (`mod.rs` + `bcrypt`, `bcrypt_sha256`,
  `pbkdf2`, `scrypt`, `plaintext`) moved from
  `crates/core/src/common/password_hashing/` to `crates/password-hashing/src/`
  (git tracks all six files as renames, `mod.rs` → `lib.rs` at 98%). Only edits:
  a crate-level doc note and a corrected path comment. **No hashing logic
  changed.**
- **`crates/core`**: removed `pub mod password_hashing;` from `common.rs` and
  dropped the now-unused `bcrypt` and `scrypt` dependencies from `Cargo.toml`
  (`argon2`, `subtle`, `hmac`, `sha2`, `base64` stay — they are used elsewhere in
  core). Core never used the module internally — it only re-exported it.
- **Consumers**: `identity-driver-sql` (4 files), `appcred-driver-sql`
  (1 file) and the `keystone` binary (dummy-hash cache reset) now depend on the
  new crate and import it directly
  (`openstack_keystone_core::common::password_hashing` →
  `openstack_keystone_password_hashing`). Call sites are unchanged — the public
  API is identical.
- **Workspace**: registered the new crate in `Cargo.toml` `members` and
  `[workspace.dependencies]`.

## Actions taken / verification

- `cargo check --workspace` — clean.
- `cargo test -p openstack-keystone-password-hashing` — 30/30 pass.
- `cargo test -p openstack-keystone-identity-driver-sql -p openstack-keystone-appcred-driver-sql` — pass.
- `cargo test -p openstack-keystone-core` — 594 pass, 0 fail.
- `cargo clippy` on the new crate — no warnings.
- **Dependency check (the point of the issue):**
  `cargo tree -p openstack-keystone-core -i bcrypt` and `... -i scrypt` both now
  report *"did not match any packages"* (previously both were in core's tree).
  The new crate correctly owns `bcrypt` and `scrypt`.

## Note on the trait

The issue suggested keeping the hasher trait in `core`. The `PasswordHasher`
trait is a private, impl-only detail shared by the hashers and their static
dispatch; splitting it from its only implementors would add an awkward seam with
no consumer benefit, so the entire self-contained module moved together and
nothing password-hashing-specific remains in `core`.


gtema (approved):
Cool, thanks

Issue #954, opened 2026-07-09
Implement domain config (per-domain identity backend driver configuration)


gtema:
# Domain Config

## Problem

Python Keystone supports per-domain identity backend configuration, allowing each domain to use a different identity driver (SQL or LDAP) with its own connection settings. Currently the Rust Keystone does not implement this feature. This is a key multi-tenant feature: it allows operators to onboard new domains with their own LDAP backends without restarting Keystone.

## Python Keystone Reference

The feature is described in two places:

- [API reference: Domain configuration](https://github.com/openstack/keystone/blob/stable/2026.1/api-ref/source/v3/domains-config-v3.inc)
- [Admin guide: Domain-specific configuration](https://github.com/openstack/keystone/blob/stable/2026.1/doc/source/admin/domain-specific-config.inc)
- [Original spec (Kilo): Domain Configuration Storage](https://wiki.openstack.org/wiki/Blueprints/Kilo/domain-config-ext)

## What It Does

1. **Per-domain identity driver selection**: Each domain can use its own identity backend driver (ldap or sql). A domain without domain config falls back to the global default driver.
2. **Per-domain LDAP options**: When a domain uses LDAP, each domain can have its own url, user_tree_dn, password, and other LDAP connection parameters.
3. **Two config sources**: Configuration can be stored in either (a) an SQL table, managed via REST API, or (b) per-domain config files on the filesystem (domain_config_dir). The SQL source takes precedence.
4. **No restart on update**: Changes via API take effect immediately without Keystone restart (cache-based).

## API Specification

Configuration is organized in groups. Only identity and ldap groups are supported.

### Groups and Options

| Group | Option | Description | Readable on GET? |
|-------|--------|-------------|-----------------|
| identity | driver | Identity backend driver name (e.g., ldap, sql) | Yes |
| ldap | url | LDAP server URL | Yes |
| ldap | user_tree_dn | Base DN for user tree | Yes |
| ldap | password | Bind password | **No** (sensitive) |
| ldap | user | Bind DN user | Yes |
| ldap | suffix | LDAP suffix | Yes |
| ldap | page_size | LDAP page size | Yes |
| ldap | use_tls | Use StartTLS | Yes |
| ldap | tls_cacertfile | TLS CA cert path | Yes |
| ldap | group_tree_dn | Base DN for group tree | Yes |
| ldap | group_mapping | LDAP-to-Keystone attribute mapping | Yes |
[...]

PR #1138, opened 2026-08-01
feat: Add domain config types and backend trait

### LINKED ISSUE #955: Phase 0: Domain config core types and backend trait
## Scope

Implement the foundational types and backend trait for domain configuration.

### `crates/core-types/` — Add `domain_config.rs`

- `DomainConfig` struct with `identity` and `ldap` groups
- `DomainConfigIdentity` with `driver: String`
- `DomainConfigLDAP` with all LDAP options (matching Python's list)
- `DomainConfigCreate`, `DomainConfigUpdate`, `DomainConfigOption` types
- Mark `password` and other sensitive fields

### `crates/core/src/backend.rs` — Add `DomainConfigBackend` trait

- CRUD methods for domain, group, and option levels
- Default config getters
- Full method signature list:
  - `create_domain_config(domain_id, config)`
  - `get_domain_config(domain_id)`
  - `get_domain_config_group(domain_id, group)`
  - `get_domain_config_option(domain_id, group, option)`
  - `update_domain_config(domain_id, config)`
  - `update_domain_config_group(domain_id, group, config)`
  - `update_domain_config_option(domain_id, group, option, value)`
  - `delete_domain_config(domain_id)`
  - `delete_domain_config_group(domain_id, group)`
  - `delete_domain_config_option(domain_id, group, option)`
  - `get_default_config()`
  - `get_default_group(group)`
  - `get_default_option(group, option)`

## References

- [API ref: domains-config-v3.inc](https://github.com/openstack/keystone/blob/stable/2026.1/api-ref/source/v3/domains-config-v3.inc)
- [openstacksdk domain_config.py](https://github.com/openstack/openstacksdk/blob/master/openstack/identity/v3/domain_config.py)

## Parent issue

Part of #954

magdang:
Phase 0 of #954: the core types and the driver interface for per-domain
identity backend configuration. No SQL driver, no API handlers, no policy
files — those are Phases 1–5 of the parent issue.

Closes #955.

## What's here

**`crates/core-types/src/domain_config/`**

| Type | Purpose |
|---|---|
| `DomainConfigGroupName` | `identity` \| `ldap`, parsed from the request path |
| `DomainConfigOption` | flat `(group, option, value, sensitive)` row a driver stores |
| `DomainConfigValue` | the option value, JSON verbatim |
| `DomainConfigIdentity` / `DomainConfigLdap` | typed groups, one field per supported option |
| `DomainConfigGroup` | one group with its options |
| `DomainConfig` | a domain's whole configuration (`{"config": {...}}`) |
| `DomainConfigCreate` / `DomainConfigUpdate` | PUT (replace) vs PATCH (merge) payloads, distinct types so they can't be swapped at a call site |
| `DomainConfigProviderError` | provider error, incl. the `InvalidDomainConfig` / `DomainConfigNotFound` cases |

**`crates/core/src/domain_config/`** — the `DomainConfigBackend` trait, with
the whole-config / group / option method triple for each verb plus the three
default-config getters, and the `DatabaseError` conversion.

## Fidelity to python-keystone

The option model is transcribed from `keystone/resource/core.py`'s
`DomainConfigManager`, not paraphrased:

- `whitelisted_options` — `identity: [driver, list_limit]` and 49 `ldap`
  options, `sensitive_options = {identity: [], ldap: [password]}`. A test
  asserts the counts and that every whitelisted option has a struct field;
  the lists were also diffed against the upstream source programmatically
  (names *and* order identical).
- Upstream spells the list out rather than deriving it from the `[ldap]`
  section, precisely so a new `[ldap]` option can't become domain-writable by
  accident. Kept that property.
- `_assert_valid_group_and_option` and `_assert_valid_config` are ported
  including their exact wording, so the eventual API returns the same
  messages (403 for an unsupported group/option, per the api-ref).
- `get_config_default` reports unset options as `null` rather than omitting
  them; `DomainConfig::default_options` reproduces that, while the typed
  `DomainConfig::defaults` omits them.
- `_list_to_config` does not re-validate on read, and neither does
  `DomainConfig::from_options` — see "Read tolerance" below.

Note: the parent issue's option table is a paraphrase and lists names that
don't exist upstream (`tree_filter`, `connection_pooling`, `user_mapping`,
`pool_startup_sleep`, …). This follows the actual Python whitelist.

## Sensitive options

`ldap.password` is the one sensitive option. It is a `secrecy::SecretString`
with `#[serde(skip_serializing)]`, following the existing
`IdentityProvider::oidc_client_secret` pattern in this crate, so:

- it can be written and used to bind, and survives a round trip through
  `to_options()` / `from_options()` (the identity backend needs it),
- it never reaches an API response — `DomainConfig` serialization drops it,
- it never reaches a log line — `SecretString` keeps it out of `Debug`, and
  `DomainConfigOption` has a hand-written `Debug` that redacts the value of
  any option flagged sensitive,
- it has no readable default: `default_option()` refuses it outright.

The **option-scoped** backend methods take and return `DomainConfigOption`
rather than a bare `DomainConfigValue` for the same reason: a value on its
own carries no sensitivity marker, so it renders in full under `Debug` and
serializes transparently. Since drivers in this workspace are conventionally
annotated with `#[tracing::instrument]`, a bare value would have put a bind
password into the span fields of every write to `ldap.password`, and the
obvious handler would have echoed it into the PATCH response body. The group
and whole-config paths were already structurally safe via `skip_serializing`;
this one needed the stronger type. It is the one place this deviates from the
signature list in #955 (`(domain_id, group, option, value)` → `(domain_id,
option)`) — same information, safe by construction rather than by driver
discipline.

Tests cover each of these.

## Value typing

Values round-trip verbatim as JSON, matching upstream's `JsonBlob` columns.
Scalars additionally accept their string spelling (`"True"`, `"10"`,
`"0.1"`, `"a,b"`) and integers `0`/`1` for booleans, reproducing
oslo.config's coercion — this is how a per-domain `keystone.<domain>.conf`
file will deliver every value in the file-based-driver phase.

Coercion goes through `serde_json::Value` rather than an untagged enum so a
rejected value gets a usable message, and the offending option is pinpointed
by retrying option by option on the error path (a group has up to 49 of them
and serde reports a failing `deserialize_with` without any field context):

```
invalid value for option page_size in group ldap: expected an integer, got `"ten"`
```

## Read tolerance

`DomainConfig::from_options` skips stored options it doesn't recognize,
logging a warning, instead of failing. Validating on read would mean a single
stale row — after a whitelist narrowing, a downgrade, or a write that slipped
past validation — leaves the whole domain configuration unreadable and its
identity backend uninitializable until an operator deleted the row by hand.
The whitelist is enforced on the write path instead (serde
`deny_unknown_fields` plus `assert_valid_group_and_option` on the path
parameters), which is also where python-keystone enforces it.

## Known deviation, flagged for later phases

This workspace's `crates/config/src/ldap.rs` parses the pooling flags as
`pool` / `auth_pool`, while python-keystone (and therefore the domain-config
API contract) spells them `use_pool` / `use_auth_pool`. This module uses the
upstream names; the mapping between the two is done in
`DomainConfigLdap::defaults` and commented there. Worth deciding in Phase 5
whether `LdapProvider` should gain serde aliases — as it stands, a
keystone.conf written for Python with `[ldap] use_pool = false` is silently
ignored by the config parser.

Relatedly, `DomainConfigIdentity::defaults` resolves `[identity] list_limit`
through the global `[DEFAULT] list_limit`, the same chain
`Config::resolve_list_limit` applies — the defaults endpoint exists to report
the value a domain inherits, so reporting only the section-local value would
say `null` while listings are really capped.

## Structural note

The issue names `crates/core/src/backend.rs`; that file doesn't exist in this
workspace — backend traits live in per-domain modules
(`crates/core/src/<domain>/backend.rs`). Followed the existing convention.

## Testing

- `cargo test -p openstack-keystone-core-types` — 164 passed (41 new)
- `cargo test -p openstack-keystone-core` — 666 passed
- `cargo clippy` on both crates — no warnings from `domain_config`
- `cargo check --workspace --all-targets` — clean apart from the
  pre-existing `crates/storage/benches/*` failures (they're behind the
  `bench_internals` feature and fail on `main` too)

### DIFF
--- a/crates/core-types/src/domain_config/config.rs
+++ b/crates/core-types/src/domain_config/config.rs
@@ -1,0 +1,1220 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+//! # Domain configuration structures
+//!
+//! A domain configuration is a set of *overrides* of the global `[identity]`
+//! and `[ldap]` config sections, so it is held as the options a domain
+//! actually set — `group -> option -> value`, values verbatim as JSON, the
+//! way the persistence layer stores them — rather than as a second set of
+//! structs mirroring those sections.
+//!
+//! What the options *mean* is defined once, by
+//! [`openstack_keystone_config::IdentityProvider`] and
+//! [`openstack_keystone_config::LdapProvider`]:
+//!
+//! - [`DomainConfig::resolve_identity`] / [`DomainConfig::resolve_ldap`]
+//!   overlay a domain's options onto the global section and hand back the
+//!   very struct the identity backend is configured with;
+//! - [`DomainConfig::defaults`] serializes the global sections and keeps the
+//!   whitelisted keys, which is what the `/v3/domains/config/…/default`
+//!   endpoints answer;
+//! - [`DomainConfig::validate_values`] checks a client's values by decoding
+//!   them into those same structs.
+//!
+//! The only thing this module states about an individual option is whether a
+//! domain may set it at all (see [`super::whitelisted_options`]) — names, not
+//! types. An option added to `LdapProvider` therefore cannot drift from its
+//! domain-configurable counterpart, because there is no counterpart.
+
+use std::borrow::Cow;
+use std::collections::hash_map::Entry;
+use std::collections::{BTreeMap, HashMap};
+use std::fmt;
+use std::ops::{Deref, DerefMut};
+
+use openstack_keystone_config::{Config, IdentityProvider, LdapProvider};
+use serde::de::DeserializeOwned;
+use serde::ser::SerializeMap;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde_json::{Map, Value};
+
+use super::lenient;
+use super::{
+    DomainConfigGroupName, DomainConfigOption, DomainConfigProviderError, DomainConfigValue,
+};
+
+/// One configuration group with the options a domain set in it.
+///
+/// Returned by the group-scoped endpoints
+/// (`/v3/domains/{domain_id}/config/{group}`), which address exactly one
+/// group.
+#[derive(Clone)]
+pub struct DomainConfigGroup {
+    /// The group these options belong to.
+    name: DomainConfigGroupName,
+    /// The options, by name.
+    options: Map<String, Value>,
+}
+
+impl DomainConfigGroup {
+    /// An empty group.
+    ///
+    /// # Parameters
+    /// - `name`: The group.
+    ///
+    /// # Returns
+    /// - `Self` - The group with no option set.
+    pub fn new(name: DomainConfigGroupName) -> Self {
+        Self {
+            name,
+            options: Map::new(),
+        }
+    }
+
+    /// Name of the group.
+    ///
+    /// # Returns
+    /// - `DomainConfigGroupName` - The group discriminant.
+    pub fn name(&self) -> DomainConfigGroupName {
+        self.name
+    }
+
+    /// The options set in the group.
+    ///
+    /// Sensitive options are included; see [`Self::serialize`] for what
+    /// reaches a response.
+    ///
+    /// # Returns
+    /// - `&Map<String, Value>` - The options by name.
+    pub fn options(&self) -> &Map<String, Value> {
+        &self.options
+    }
+
+    /// Read a single option.
+    ///
+    /// # Parameters
+    /// - `option`: The option name.
+    ///
+    /// # Returns
+    /// - `Option<&Value>` - The value when the domain set the option.
+    pub fn get(&self, option: &str) -> Option<&Value> {
+        self.options.get(option)
+    }
+
+    /// Set a single option.
+    ///
+    /// # Parameters
+    /// - `option`: The option name.
+    /// - `value`: The value to store.
+    ///
+    /// # Returns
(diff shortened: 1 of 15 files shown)


gtema (changes requested):
I would rather reuse the existing config:LdapProvider struct rather than duplicate it - a lot of params repeated and very easy to deviate. 
The python code goes a way of whitelisting which particular config options can be set per group. In rust it is not that easy to do it the same way either than passing LdapConfigProvider and IdentityProvider config structs and sanitize them by serializing them into json Value and stripping all keys not in the whitelisted list. In the database we anyway store the json serialized values, so it could be used. This way we would not need to repeat the structs and eliminate the danger of drifting (missing parameters of applying different type conversions to them)

crates/core-types/src/domain_config/config.rs
```diff
@@ -0,0 +1,1080 @@
...
+/// Fields correspond 1:1 to python-keystone's `whitelisted_options['ldap']`
+/// plus the single sensitive option, `password`. The list is intentionally
+/// spelled out rather than derived from the `[ldap]` config section: a new
+/// `[ldap]` option must never become domain-writable by accident.
+///
+/// `PartialEq` is intentionally not derived: `password` is wrapped in
+/// [`SecretString`], which does not implement `PartialEq` by design.
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct DomainConfigLdap {
```
gtema: with Value based approach I actually wanted to get rid of this DomainConfigLdap struct which duplicates all properties already defined in the config crate. I suggest passing the config struct around (this is actually a general idea of overriding config per domain) and sanitize it during serialization/deserialization

magdang: Done in `cd103070`. `DomainConfigLdap` and `DomainConfigIdentity` have been removed. Domain overrides remain sparse `Value` maps so PATCH and row storage preserve only explicitly configured keys. Serialization filters `IdentityProvider` and `LdapProvider` to the explicit whitelist and omits the LDAP password; resolution overlays the stored values and deserializes back into those config structs. I also kept substituted LDAP providers in a Debug-redacting wrapper and made an explicit null password suppress global-password inheritance. This keeps the config crate as the single source of option names, types, coercions, and defaults.

gtema (approved):

PR #1153, opened 2026-08-05
test(api): Cover v4 handlers and fix role assignment routing

ShJ-code:
## Summary

Closes the v4 half of #993. The v3 half landed in #1084; this finishes the
issue.

Adds 129 live-server tests across the seven remaining endpoint families,
each handler getting the three cases CLAUDE.md mandates — authorized 2xx,
policy denial 403, invalid token 401:

| suite | tests |
|---|---|
| `api_v4/identity/user` | 19 |
| `api_v4/role` | 17 |
| `api_v4/role_assignment` | 29 |
| `api_v4/auth/token` | 11 |
| `api_v4/api_key` | 21 |
| `api_v4/token_restriction` | 15 |
| `api_v4/federation/identity_provider` | 17 |

This also retires the last unreferenced helper modules (`api_key`,
`token_restriction`, and `federation::identity_provider`, which was
previously reachable only from the SCIM suites), satisfying the issue's
"0 unreferenced helper modules" criterion.

**Deviation from the issue text:** #993 lists federation identity providers
as a v3 hole, but `federation` is nested only by `api::v4` — the v3 router
has no federation entry — so that suite lives under `tests/api_v4/`.

### Two production bugs the tests found

**`fix(api): Merge v4 role assignment routes`** — the role assignment
sub-routers register absolute paths (`/role_assignments`,
`/projects/{project_id}/users/…`, `/system/users/…`), so `api::v3` merges
them. `api::v4` nested them under `/role_assignments`, which prefixed those
absolute paths: `/v4/role_assignments/role_assignments` was published while
the intended `/v4/role_assignments` and `/v4/projects/…` were unrouted.
Nothing referenced the doubled paths. Switched to `.merge()` and pinned the
shape with `api::tests::v4_role_assignment_routes_are_merged_not_nested`,
which asserts the intended paths do not 404 and the nesting artefacts do.

**`fix(token): Run restriction update inside its transaction`** — `update`
opened a transaction and then issued every statement against the pool. On a
single-connection pool the transaction held the only connection while the
update asked for a second, so `PUT /v4/tokens/restrictions/{id}` blocked for
the full acquire timeout and returned 500. The transaction also wrapped
nothing: the row update and the role association insert/delete
autocommitted individually, so a mid-way failure left a restriction and its
roles inconsistent. The existing `test_update` cannot catch either —
`MockDatabase` logs statements in issue order whether or not they went
through the transaction — so the regression test runs against a real
single-connection SQLite pool.

### Test-side notes

Helper modules use the #992 `crud_endpoint!` macro rather than hand-written
`RestEndpoint` impls (`token_restriction` 252→71 lines,
`federation::identity_provider` 296→142, and the user-groups sub-resource
via the macro's `parent` form). Two arms were added to express v4:
`update_put`, because every v4 update handler declares `put` where its v3
counterpart declares `patch`, and `delete_fn`, because `v3::user::User`
already carries the `DeletableResource` impl. `api_key` stays hand-written
and says why: every operation carries a mandatory `domain_id` query
parameter, which the macro's show/update/delete arms do not model.

Negative assertions go through the #992 `assert_forbidden` /
`assert_unauthorized` helpers so a wrong status reports the full error
chain, matching the merged v3 suites.

Every 2xx here is `admin`, and each suite documents why: the non-admin
allow rules pair a role with a **domain** scope, and no
`PUT /v3/domains/{domain_id}/users/{user_id}/roles/{role_id}` handler
exists, so no real user can hold `manager`/`reader` on a domain scope. The
403 fixtures are therefore real project-scoped role holders — they *hold*
the role and are still denied, which proves the domain-scope gate rather
than mere role absence. The few non-admin allow rules that *are* reachable
(system-scoped `reader` on user list and project role check, system
`manager` on system grants, unfiltered IDP list) get genuine
underprivileged 2xx cases.

## Test plan

All commands run locally against a live server via `tools/start-api.sh`.

| check | result |
|---|---|
| `cargo nextest run -p test_api --profile ci-api --test integration_api_v3 --test integration_api_v4` | 310/310 pass |
| `cargo nextest run -p test_api --profile ci-api --test scim_v2` | 64/64 pass |
| `cargo nextest run --all-features --profile ci` | 3021 pass, 2 skipped |
| `cargo nextest run -p test_integration --profile ci` | 229 pass, 106 skipped |
| `opa test policy` / `opa fmt policy --check-result` | 346/346, clean |
| `cargo fmt -- --check` | clean |
| `cargo test --doc` | pass |
| `--dump-openapi yaml` | generates; confirms `/v4/role_assignments`, `/v4/projects/…/roles` and `/v4/system/…/roles` are published and zero nesting artefacts remain |
| `tools/check_{policy_handler_coverage,rego_undefined_argument_footgun,security_checklist_sast,event_payload_no_secret_fields,delegated_policy_scope_drift_tests}.py` | all pass |

Every commit in the series was also verified to compile on its own
(`git rebase --exec 'cargo check --workspace --lib --tests'`).

## Security review checklist

Retained rather than deleted because the diff makes role-grant handlers
reachable at a new prefix and changes a token-restriction write path. No
authorization decision, policy input, scope shape or auth method is
touched: outside `tests/`, the diff is three files —
`api/mod.rs` (test-only), `api/v4/mod.rs` (`.nest()` → `.merge()`), and the
token restriction SQL driver.

      (`authentication_context()`, delegation object)? (I1/I2) — No
      authorization decision is added or modified.
- [x] New delegation-sensitive policy rule? ... (I2/I3) — No `policy/`
      file is touched.
- [x] New scope shape or redemption path for a delegated auth? ... (I4) —
      None.
- [x] New `ScopeInfo` variant or auth method? ... (I5, Gate J) — None; Gate
      J passes.
- [x] New lookup by a client-derivable key (like `sha256(access)`)? ...
      (I6) — None.
      Gate I) — Nothing new reaches OPA; Gate I passes.
- [x] New list/collection endpoint? ... (I8) — No new handler. The v4
      route merge exposes existing v3 handlers, which keep their existing
      per-item enforcement, at the `/v4` prefix.
      request-supplied scope? (I5) — No.
      positive tests proving the happy path works? — Yes: every handler
      has a 403 case driven by a real underprivileged user through the
      live auth path, plus a 401 case.
      end-to-end, not just the inner helper? — These are live-server tests;
      every request goes through the real auth middleware end to end.


ShJ-code:
The failing test might need a rerun, which I am not authorized to do currently.

gtema (approved):

PR #1154, opened 2026-08-06
fix(policy): Bind credential listing to own user

### LINKED ISSUE #1117: Invalid policy for listing credentials
Python Keystone has a following policy for listing credentials: `"identity:list_credentials": "(rule:admin_required) or (role:reader and system_scope:all) or user_id:%(target.credential.user_id)s"`. On our side we properly allow admin, system reader listing credentials and anybody with the member role missing enforcing that the regular user would be only allowed to see own credentials. While the "credential.show" policy would still block the credentials from being returned it enables regular user to perform a DDoS attack by simply scanning through all credentials in the system.
The listing policy need to be modified replacing the "member" role with the "credential.user_id not null and match user's own user_id"

ShJ-code:
## Summary

Closes #1117.

`identity/credential/list` granted any caller holding `member` an unfiltered
list of every credential in the system.

Nothing leaked: the per-item `identity/credential/show` re-check (I8,
CVE-2019-19687) still dropped the rows the caller could not read. But the rows
were fetched and OPA-evaluated *before* being discarded, so one cheap request
became a full-collection scan plus one policy round-trip per row — a
resource-exhaustion amplifier available to any authenticated user.

The `member` rule is replaced by an ownership rule: the request must carry a
`user_id` filter naming the caller. `admin` and system-scoped `reader` keep the
unfiltered collection view. This mirrors python keystone's
`identity:list_credentials`:

```
(rule:admin_required) or (role:reader and system_scope:all)
  or user_id:%(target.credential.user_id)s
```

whose `user_id:%(target.credential.user_id)s` likewise resolves against the
query parameters and so denies an unfiltered list to an unprivileged caller. It
also matches `identity/user/application_credential/list`, which already gates
the sibling collection on the caller's own `user_id`.

**Ownership alone is the grant; no role is required.**
`identity/credential/show` already gives a non-delegated caller their own
credential on ownership alone, and python's rule carries no role component, so
demanding `member` would deny a reader-scoped caller their own credentials for
no security gain.

**The gate is not decorative.** The filter the policy approves and the filter
the driver applies are the same value — both read `user_id` off one parsed
`CredentialListParameters` — so a request cannot be approved on a filter the
query then ignores, which would leave the scan in place behind a check that
does nothing.

**The delegation boundary is unmoved.** OSSA-2026-015 stays where it was, in
the per-item re-check. This rule sees only query parameters, so I2's
`bound_to_own_delegation_project` has nothing to anchor on; a delegated caller
filtering to their own `user_id` passes here and still has every cross-project
and unscoped record dropped downstream. `show`'s delegated branch is unchanged.

Recorded as invariant **I8a** in the security model and in ADR 0019, with a
reviewer checklist entry: I8 makes a permissive collection filter *safe*, not
*cheap*, and the two properties need separate checks. The other I8 endpoint,
`/v3/policies`, is admin/system-reader only and already satisfies I8a.

### ⚠️ Behaviour change

`GET /v3/credentials` without a `user_id` filter now returns **403** for callers
who are neither `admin` nor system-scoped `reader`; they must pass
`?user_id=<their own id>`. Previously any `member` received a (correctly
filtered, often empty) 200. Python keystone already behaves this way, so this
closes a compatibility gap rather than opening one.

Based on `main` at #1127 (issue #1035), which is already merged — this is a
single commit against `main` with no stacked dependencies.

## Test plan

| check | result |
|---|---|
| `opa test policy/` | **366/366 pass** — 6 new/extended cases in `policy/credential/list_test.rego` |
| `cargo test -p openstack-keystone --lib` | **919/919 pass** |
| `cargo test -p openstack-keystone --lib api::v3::credential` | **39/39 pass** — 4 new handler tests |
| `cargo check -p test_api --tests` | clean — 3 new/rewritten live-API tests |
| `cargo fmt --all -- --check`, `opa fmt --diff` | clean |
| `cargo clippy --no-deps` on touched crates | no new diagnostics |

Repository security gates, all pass:
`check_rego_undefined_argument_footgun.py` (Gate E),
`check_policy_handler_coverage.py` (B1),
`check_security_checklist_sast.py` (Gate J),
`check_delegated_policy_scope_drift_tests.py` (Gate C),
`check_event_payload_no_secret_fields.py` (Gate I).

**Rego** (`policy/credential/list_test.rego`) — every shape of "no usable own-user
filter" denies: key absent, `null` (the shape the handler actually emits for an
absent `?user_id=`, since the `Option` is serialized rather than skipped),
`target.credential` missing, `target` null outright, `""` from a bare
`?user_id=`, and caller `user_id` missing. Plus: another user's id denies, a
non-system `reader` gets no exemption, `system` set to anything but `all` gets
no exemption, privileged listers keep the unfiltered view, and the violation
message fires exactly once on denial and not at all on allow.

**Handler** (`crates/keystone/src/api/v3/credential/list.rs`) — four tests driven
through a **real `opa run` subprocess**, since the mocked enforcer answers
`allow` unconditionally and cannot tell the old rule from the new one:

- `unfiltered_list_by_ordinary_member_is_denied_by_real_policy` — the backend
  mock deliberately sets **no expectation** for `list_credentials`, so `mockall`
  fails the test if the handler reaches the driver at all. That is the DoS
  property under test: a 403 that still scanned the table would fix nothing.
- `list_filtered_to_another_user_is_denied_by_real_policy` — same, for
  `?user_id=<someone else>`.
- `own_filtered_list_by_ordinary_member_is_allowed_by_real_policy` — asserts the
  filter reaches the driver via `withf`, proving the query is narrowed in SQL.
- `unfiltered_list_by_system_reader_is_allowed_by_real_policy` — privileged
  listers keep their cross-user view.

The pre-existing delegation test now sends `?user_id=u1`, which is load-bearing:
the caller holds only `member`, so an unfiltered list is denied outright by the
collection check and would never reach the per-item pass that test exists to
exercise.

**Live API** (`tests/api/tests/api_v3/credential/list.rs`) — the suite's ambient
session is the bootstrap admin, which takes the privileged branch and cannot
observe the restriction, so these act through a genuinely unprivileged
project-scoped `member` built with the shared `ProjectScopedUser` /
`two_members` / `cleanup_project_scoped_users` fixtures: own-`user_id` allowed,
unfiltered forbidden, `?type=totp` alone forbidden (a type filter is not a way
around the missing ownership filter), another user's id forbidden. Compile-
checked here; not executed, as the live suite needs the SPIRE + OPA harness.

## Security review checklist

      (`authentication_context()`, delegation object)? (I1/I2) — no; the new
      branch reads only `input.credentials.user_id` (from the chain) and a
      query parameter. No scope field is consulted.
- [x] New delegation-sensitive policy rule? Is the fact it needs projected
      onto `Credentials` from the chain, and does the rule anchor on
      `delegated_project_id` + carry the scope-drift tripwire? (I2/I3) — the
      new branch is deliberately **not** delegation-sensitive: it sees only
      query parameters, so there is no resource project to anchor on. The
      OSSA-2026-015 boundary remains entirely in the per-item
      `identity/credential/show` re-check, which is unchanged, and Gate C
      confirms its scope-drift negative case still exists.
- [x] New scope shape or redemption path for a delegated auth? Are effective
      roles still bounded by the delegation? (I4) — n/a, no new scope shape or
      redemption path.
- [x] New `ScopeInfo` variant or auth method? (I5, Gate J) — n/a; Gate J passes.
- [x] New lookup by a client-derivable key (like `sha256(access)`)? (I6) — n/a.
      Gate I) — no. The policy target is unchanged (`{"credential": <query
      params>}`, i.e. `type` and `user_id` only); no `blob` or `encrypted_blob`
      is involved. Gate I passes.
- [x] New list/collection endpoint? Does it re-check each item with the
      per-item read policy? (I8) — not new, and the per-item re-check is
      preserved verbatim, including propagating non-`Forbidden` policy errors
      rather than silently truncating.
- [x] Same endpoint: can an unprivileged caller request the whole collection
      and leave the per-item pass to sort it out? Is the value the collection
      policy gates on the same one the driver actually filters by? (I8a) —
      this is the bug being fixed; both read one parsed
      `CredentialListParameters`, and the handler tests assert the driver is
      never reached on denial.
      request-supplied scope? (I5) — no; the change only narrows.
      positive tests proving the happy path works? — yes: 3 of the 4 new
      handler tests and 2 of the 3 new live-API tests are negative, plus the
      Rego denial matrix above.
      end-to-end, not just the inner helper in isolation? — the live-API tests
      drive the full password-auth path against a real server. The handler
      tests inject `ValidatedSecurityContext` fixtures from
      `real_policy_fixtures`, which is appropriate here: this PR changes no
      role calculation or scope validation, only the Rego decision, and those
      tests exercise it through a real `opa run`.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -37,6 +37,10 @@ behind each item.
       Gate I)
 - [ ] New list/collection endpoint? Does it re-check each item with the
       per-item read policy? (I8)
+- [ ] Same endpoint: can an unprivileged caller request the whole collection
+      and leave the per-item pass to sort it out? Is the value the
+      collection policy gates on the same one the driver actually filters
+      by? (I8a)
 - [ ] Does the change let a narrow auth method be broadened by a
       request-supplied scope? (I5)
 - [ ] Are there negative tests proving the escape is blocked, not just
--- a/crates/keystone/src/api/v3/credential/list.rs
+++ b/crates/keystone/src/api/v3/credential/list.rs
@@ -41,6 +41,15 @@ use openstack_keystone_core::policy::PolicyError;
 /// Pagination is applied *after* this per-item filtering, over the
 /// already-policy-approved set, matching the over-fetch-by-one convention
 /// used everywhere else.
+///
+/// The first phase is what keeps the second one cheap: unless the caller is
+/// a privileged lister (`admin`, or `reader` on the system scope),
+/// `identity/credential/list` demands a `user_id` filter naming the caller
+/// (issue #1117), so the driver narrows the query in SQL rather than
+/// handing the whole table to the per-item loop. The filter the policy
+/// approves and the filter the driver applies are the *same* value — both
+/// read `CredentialListParameters::user_id` off this one parsed struct — so
+/// the two cannot diverge.
 #[utoipa::path(
     get,
     path = "/",
@@ -323,13 +332,19 @@ mod tests {
     /// Gate B3 (security review V3a, issue #979): the per-item re-check
     /// (ADR 0019 §2, CVE-2019-19687) is where `list`'s delegation boundary
     /// (OSSA-2026-015) actually lives -- `identity/credential/list.rego`
-    /// itself lets any member attempt to list, so this drives a delegated
-    /// caller's list through the real `identity/credential/show.rego`
-    /// decision (via a real `opa run` subprocess) over a mixed batch: one
-    /// record bound to the delegation's own project (must survive), one
-    /// bound to a different project, and one unscoped (both must be
-    /// silently dropped, never surfaced as an error). Requires `opa` on
-    /// `PATH`.
+    /// has no resource project to anchor on, so it admits a delegated
+    /// caller filtering to their own `user_id` and leaves the project
+    /// boundary entirely to the re-check. This drives such a caller's list
+    /// through the real `identity/credential/show.rego` decision (via a
+    /// real `opa run` subprocess) over a mixed batch: one record bound to
+    /// the delegation's own project (must survive), one bound to a
+    /// different project, and one unscoped (both must be silently dropped,
+    /// never surfaced as an error). Requires `opa` on `PATH`.
+    ///
+    /// The `?user_id=u1` filter is load-bearing since issue #1117: the
+    /// caller holds only `member`, so an unfiltered list is now denied
+    /// outright by the collection-level check and would never reach the
+    /// per-item pass this test exists to exercise.
     #[tokio::test]
     async fn delegated_caller_list_is_filtered_to_own_project_by_real_policy() {
         let mut credential_mock = MockCredentialProvider::default();
@@ -375,7 +390,7 @@ mod tests {
             .as_service()
             .oneshot(
                 Request::builder()
-                    .uri("/")
+                    .uri("/?user_id=u1")
                     .extension(vsc)
                     .body(Body::empty())
                     .unwrap(),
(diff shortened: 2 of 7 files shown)


gtema (approved):
great work

PR #1192, opened 2026-09-02
feat: Add vendor data JWT API for SPIRE integration (Phase 2)

gtema:
Implements Phase 2 of doc/plans/spire-integration.md: POST
/v4/vendordata signs a short-lived attestation JWT for a booting VM
instance, using a dedicated per-domain signing key wholly independent
from the OAuth2 access-token key, published at its own
GET /v4/spiffe/{domain_id}/jwks endpoint.

Key pieces:
- openstack_keystone_core::spiffe_key: filesystem-backed attestation
  signing key provider (own Primary/Previous/Pending rotation via the
  existing generic key-repository, no OAuth2 coupling).
- openstack_keystone_core::nova_client: Nova ownership-verification
  provider (Fix 1) with a trait-in-core/reqwest-impl-in-keystone split
  matching the existing K8sHttpClient pattern. Fails closed (503) on
  a Nova-side error, 403 on an explicit ownership mismatch.
- crates/keystone/src/api/v4/vendordata.rs: the POST handler, gated to
  the internal SPIFFE interface, deriving the calling compute host
  directly from the request's SpiffeId extension.
- crates/keystone/src/api/v4/spiffe.rs: the dedicated attestation JWKS
  endpoint.
- policy/vendordata/create.rego, doc/src/adr/0032-vendor-data-jwt.md.

Known deviations from the plan doc, detailed in the ADR: the
attestation key is filesystem- rather than Raft-backed (single-node
only, matching the Phase 0 JWS provider's own posture); the Nova
ownership check uses a static operator-provisioned admin token rather
than a dynamically minted service credential; and the compute-host
claim is derived directly from the SpiffeId extension rather than
through Phase 3's generic SPIFFE claim-flattening in the mapping
engine.

Assisted-by: Claude <noreply@anthropic.com>
Signed-off-by: Artem Goncharov <artem.goncharov@gmail.com>


crates/config/src/vendordata.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// ownership cross-check (`server.hypervisor_hostname` is gated behind
+    /// `os-extended-server-attributes`, normally admin-only).
+    ///
+    /// This is a known simplification for Phase 2: a static, operator
+    /// provisioned token rather than a dynamically minted, auto-refreshed
+    /// system-scoped service credential. Treat it with the same operational
+    /// rigor as a signing key (see the plan's Security Considerations
+    /// table) and rotate it manually until a follow-up phase automates
+    /// credential issuance.
+    #[serde(default)]
```
gtema: should be done differently

Issue #1298, opened 2026-09-18
raft: each node bootstraps its own random DEK; snapshots between nodes and backup restore on a fresh cluster cannot be decrypted; TPM KEK cannot be shared


gtema:
Part of #1290 / #1288. Severity: **critical (blocks adding nodes after compaction and blocks disaster recovery)**.

## What

`bootstrap_dek` generates a fresh random DEK on first boot of every node and persists it wrapped under that node's KEK in that node's Fjall `meta` (`crates/storage/src/lib.rs:471-482`). Nothing distributes it: the DEK only becomes cluster-wide after the first `InstallDek` (i.e. the first `keystone-manage storage rotate-dek`). Until then node 1, 2 and 3 each encrypt their Fjall state under three different `StateDek`s. This "works" for log replication only because Raft log entries carry the *plaintext* command and each node re-encrypts locally.

## Consequences

1. **Snapshot install fails.** A follower that receives a snapshot (learner catch-up after the default 5000-entry compaction, or a lagging node) gets ciphertext produced under the leader's `StateDek`. Its own `DekEpoch` has the same `version` (1) but different key bytes, so `decrypt_state_by_version` fails GCM verification on every record, three failures quarantine the partition, and the node serves nothing (`state_machine.rs:756-801`). Combined with #1293 (metadata not in the snapshot) the `dek_version` hint is absent, so the legacy probe path fails the same way.
2. **Restore is impossible on a fresh cluster.** `distributed.md` §Backup and Restore and `RestoreCommand`'s help say: bootstrap a fresh single-node cluster, then `restore`. The fresh node has a new random DEK; `decrypt_snapshot_file` requires an epoch with matching `version` *and* key bytes (`state_machine.rs:1223-1247`), so the restore returns "no DEK epoch matching snapshot version 1". ADR 0016-v2 §7 step 3 ("re-wrapping the restored DEK under the new cluster's runtime KEK") and the "DEK manifest in the backup bundle" are not implemented — the backup has no way to carry key material at all.
[...]

PR #1330, opened 2026-09-19
fix(storage): Fix full_snapshot RPC deadlock

### LINKED ISSUE #1329: raft: leader never resumes replicating to a newly added learner once its log has been purged
Part of #1289 / #1288. Found while implementing #1298. Severity: **high, needs confirmation in a real multi-process deployment** (see "Open questions" — this may be a genuine openraft/app replication stall, or an artifact of `test_cluster.rs`'s single-process, multiple-tokio-runtime test harness; either way it currently blocks writing the integration test #1289's own exit criteria and #1293's "Proposed fix" ask for).

## What

In `crates/storage/tests/test_cluster.rs`, a single-node cluster whose log has been snapshotted and purged (`raft.trigger().snapshot()` then `raft.trigger().purge_log(upto)`, both confirmed complete via `RaftMetrics.snapshot`/`RaftMetrics.purged`) never replicates anything to a learner added afterward. The learner's own `RaftMetrics.last_log_index` stays `None` indefinitely — not "slow", but never advances even after 10s of polling (multiple heartbeat/election-timeout windows at the default `heartbeat_interval = 500`, `election_timeout_min/max = 1500/3000`).

This reproduces identically via two independent code paths, ruling out anything specific to #1298's new `Storage::join_cluster`/`FetchDek` code:

1. A raw `admin_client.add_learner(AddLearnerRequest { node: Some(new_node(2)) })` call against the leader's `ClusterAdminService` (the same pattern `test_cluster_inner` already uses successfully for a *pre-purge* join).
2. `Storage::join_cluster` (added in #1298), which also calls `add_learner` after fetching the DEK.

`add_learner`'s `blocking: true` wait (openraft's `check_replication_upto_date`, `raft/api/management.rs`) returns quickly in both cases and gives false confidence: for a log this short, `distance <= replication_lag_threshold` is satisfied trivially regardless of whether any entry actually reached the learner, so `add_learner` returning `Ok` is not evidence of real catch-up.

## Reproduction

In `test_cluster.rs`'s existing harness (`InstanceHolder`, `start_raft_app`, `new_admin_client`):

1. Bring up node 1 alone, `Init` as a single-node cluster.
2. Write ~20 keys via `StorageApi::set_value`.
3. `instance1.storage.raft.trigger().snapshot().await?;` — wait until `metrics().snapshot` is `Some`.
4. `instance1.storage.raft.trigger().purge_log(upto).await?;` — wait until `metrics().purged` is `Some(LogId { index: upto, .. })`.
5. Bring up node 2, wait for its own listener to bind.
6. Call `add_learner` for node 2 against node 1's admin service (or `Storage::join_cluster`).
7. Poll `instance2.storage.last_log_index()` for up to 10s — it never leaves `None`.

(This is exactly the setup #1298's own regression test originally tried to use, per its commit history — it was rewritten to join *before* any purge specifically to route around this bug, at the cost of not exercising the "join after compaction" scenario #1298's issue text and #1289's exit criteria both call out.)

## What I did *not* establish

I bisected far enough to be confident this is unrelated to DEK/KEK correctness, but not far enough to pin the failure on a specific layer:

- Whether the leader's `RaftNetwork`/`NetworkManager` is even attempting to open a replication connection to the new learner after the purge (no tracing output was visible in the test process to confirm either way — the leader and learner each run in a separate OS thread with its own `AsyncRuntimeOf::<TypeConfig>::new(1)` tokio runtime inside *one* process, and `tracing_test::traced_test`'s subscriber did not appear to capture events from those other-thread runtimes).
- Whether this is an openraft library behavior around a leader whose log is now empty (no entries at or after the purge point to replicate, so nothing to send until the *next* write), an app-level network/config issue specific to this repo's `NetworkManager`, or purely an artifact of running multiple "nodes" as threads sharing one process instead of separate processes/containers.

## Suggested next steps

- Reproduce with real separate processes (or at minimum separate Tokio runtimes proven not to share any global state) before concluding this is a production bug rather than a test-harness limitation.
- If confirmed: instrument/step through openraft's replication-stream startup for a newly added learner when the leader's log is already empty at purge time — does it wait for the *next* write to notice the learner needs a snapshot?
- Add the integration test #1293 already asks for once this is fixed: 3-node cluster, write past the snapshot policy threshold across multiple keyspaces, purge, join a 4th node, assert every read matches the leader.

## Related

- #1298 (found while writing this issue's repro as that fix's regression test)
- #1293 (its own "Proposed fix" asks for the same "purge, join node 4, assert reads" integration test this bug currently blocks)
- #1289 (M1 exit criteria: "a node added via `storage join` after log compaction... serves the same reads as the leader")

gtema:
## Summary

Fixes #1329: a purged-log leader never resumed replicating to a newly
added learner.

Root cause: `NetworkConnection::full_snapshot` (`crates/storage/src/network.rs`)
awaited the client-streaming `Snapshot` RPC's response (`client.snapshot(rx).await?`)
*before* sending anything into the `mpsc` channel backing its request
stream (`rx`). The server side (`RaftServiceImpl::snapshot`,
`crates/storage/src/grpc/raft_service.rs`) only replies once it has read
the entire request stream. Since both the "await the response" and
"produce the stream" code lived in the same task, in that order, the
task deadlocked against itself on every attempt to send a snapshot —
silently, since the surrounding retry loop in openraft's
`SnapshotTransmitter` only logs on `Err`, and a hung `.await` produces
neither an `Err` nor any further tracing.

This is why a learner added after the leader's log was snapshotted and
purged never caught up: openraft correctly detects the need for a
snapshot and issues `Command::ReplicateSnapshot`, but the actual RPC
call to transmit it then hung forever, so `RaftMetrics.last_log_index`
on the learner stayed `None` indefinitely. This reproduced identically
with real, separate OS threads/runtimes per node, ruling out the
test-harness explanation the issue raised as an open question — it's a
deterministic bug in the RPC client code, not a threading/harness
artifact.

The fix: since the whole snapshot already lives in memory
(`Vec<u8>`) by the time `full_snapshot` runs, build the complete
request stream eagerly with `futures::stream::iter(...)` instead of
funneling it through a channel fed by code that comes after the
`.await`. This removes the self-deadlock entirely and needs no
separate feeder task.

Also updates a stale doc comment on `test_join_adopts_cluster_dek_inner`
that described this bug (discovered while working on #1298) as an
open, unfixed gap.

## Test plan

- Added `test_purge_then_join_learner_catches_up_via_snapshot` to
  `crates/storage/tests/test_cluster.rs`: brings up a single-node
  cluster, writes 20 keys, triggers `snapshot()` + `purge_log()`,
  brings up a fresh node, adds it as a learner via a raw `add_learner`
  call, and asserts it catches up via `InstallSnapshot` within 10s.
  Confirmed this test hangs forever against the pre-fix code and
  passes reliably after the fix.
- `cargo test -p openstack-keystone-distributed-storage --test
  test_cluster -- --test-threads=1`: all 13 tests pass.
- `cargo test -p openstack-keystone-distributed-storage --lib`: all 95
  unit tests pass.
- `cargo clippy -p openstack-keystone-distributed-storage --lib
  --tests` and `cargo fmt -p openstack-keystone-distributed-storage --
  --check`: clean.
- `pre-commit run --files crates/storage/src/network.rs
  crates/storage/tests/test_cluster.rs` and the `committed` commit-msg
  hook: both pass.


gtema:
The merge-queue run was dequeued for `CI_FAILURE`. Looked at [the failing "CI" run](https://github.com/openstack-experimental/keystone/actions/runs/35493140847/job/106031426505): the `Test (x86_64-unknown-linux-gnu)` job failed at the `Run REST API SCIMv2 tests` step, specifically in the `tools/start-api.sh` nextest setup script — it printed `Keystone server is started and running in the background!` then 12s later `Server failed to start.` and timed out (`SETUP FAIL [16.260s]`).

This isn't this PR's diff: it only touches `crates/storage/src/network.rs` (the raft peer-to-peer gRPC client) and a raft-only integration test, neither exercised by the SQL-backed `test_api`/SCIMv2 suite. The preceding `Run REST API tests` step in the *same* job (which also boots a live keystone server) passed a few seconds earlier, and this exact job (`Test (x86_64-unknown-linux-gnu)`) already passed on this same commit when CI ran directly against the PR branch (not through the merge queue) — see the [green run](https://github.com/openstack-experimental/keystone/actions/runs/35459211008/job/105939952578). The pattern (old server process killed, new one started immediately after, health check timing out on a fresh bind) looks like a runner-timing flake rather than a real regression.

I've kicked off a re-run of the failed job to confirm before asking for a re-queue: https://github.com/openstack-experimental/keystone/actions/runs/35493140847

---
_Generated by [Claude Code](https://claude.ai/code)_

